PR #2, opened 2021-03-04
quay as a utility registry

thomasmckay:


alecmerdler (changes requested):
@thomasmckay @jonathankingfc We should complete this proposal ASAP since the work is already under way.

PR #3, opened 2021-04-09
Add redis cache proposal

syed:


BillDett:
Sounds like simple, non-sharded HA Redis is the way to go (and avoids any big client changes).  Will be very interesting to see what effect the network has on this over the intra-Pod memcache.  We might observe an overall increase in query latency while database traffic itself goes down.

enhancements/redis-cache.md
```diff
@@ -0,0 +1,204 @@
...
+---
+
+# redis-cache-provider 
+
+Quay has a data caching mechanism which it uses to cache various objects.
+This proposal is to add Redis as another provider/implementation for this cache.
+
+`-`.
+
+To get started with this template:
```
alecmerdler, line 22: Can you remove this part of the template?

alecmerdler, line 22: Can you remove this part of the template?

alecmerdler (changes requested):
Proposal looks good, just remove the sections of text left over from the template.

PR #6, opened 2021-05-11
WIP: Initial Quay quota proposal

kleesc:


BillDett:
I think we should split out the design into database specific concerns (e.g. adding a size column) vs. architectural concerns (e.g. generating regular quota reports from a read replica instead of the read/write database).  The primary users here are Enterprise customers and it can/should dictate a more advanced database set up (say with a read replica) for installs beyond a simple department/team.  I'd also suggest that for quay.io this sort of a feature should be a paid opt-in feature instead of something we'd just add for everyone by default (this would further minimize the impact against our resources).

HammerMeetNail:
@kleesc Can we close this out in favor of https://github.com/quay/enhancements/pull/12 ?

PR #7, opened 2021-05-20
automate-setup

thomasmckay:


automate-setup/automate-setup.md
```diff
@@ -0,0 +1,87 @@
...
+
+### API
+
+`/api/v1/user/initialize'
+
+Will create a new user iff there are no users in the database.
+
+Input `name`, `password`, `email`, and `access_token`. The first three are as described in regular user create API. The `access_token` is unique in that it is the scope of permissions to create an access token for this user.
+
+The scope may be one or more of the following in a single string separated by spaces:
```
syed, line 25: Is there a need for `scope`? As I understand the idea is to initialize the `admin` user which can later be used to initialize other users. If someone creates a user with `user:read` scope only then you'll have to manually create the admin user.

thomasmckay: I originally just had `access_token: true` which would then have all permissions added to the token. I'm can't think of a reason not to give all permissions.

automate-setup/automate-setup.md
```diff
@@ -0,0 +1,87 @@
...
+
+A successful call will return the created user and credentials.
+
+For example:
+```
+{
+    "username": "admin",
+    "email": "admin@quay.example.org",
+    "encrypted_password": "QmPdFZdvfjA0JvnNPEdIZhtKzBjQmXgqQvabEXDYNL0ucHmUXhlPEUzgMThMEN6V",
+    "access_token": "6WQPRJZ5HHC6G8L2T5TM9MCGBSIX0AFJXKVVDYOH",
```
kleesc, line 75: Is this an oauth token. How is this going to be generated? You would need to also create an organization with an application to do so. Afaik you need both in order to generate an application token that would give you access to _Quay's_ api.

thomasmckay: Yes, the user is the organization. We don't enable this through quay api right now, but I've confirmed it works so no org needs to be created.

PR #10, opened 2021-11-01
[WIP] volume: Volume Size Override Proposal (PROJQUAY-1090)

jonathankingfc:
- Volume Size Override Proposal

### DIFF
--- a/enhancements/volume-overrides.md
+++ b/enhancements/volume-overrides.md
@@ -1,0 +1,190 @@
+---
+title: Neat-Enhancement-Idea
+authors:
+  - "@jonathankingfc"
+reviewers:
+  - "@flavianmissi"
+approvers:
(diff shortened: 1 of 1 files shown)


ricardomaraschini (changes requested):
@jonathankingfc I think that we gonna need more details into this Enhancement Proposal before we move on. It would be vital to have here a proposal for the actual implementation (it does not need to be the right implementation, but one that should work and that we can start working on). You gonna need to answer here questions as: 

1. How will the CRD spec look like ?
2. How will the CRD status look like (when users enter override values and when they don't)  ?
3. What is the expected behavior for the operator when an user asks for shrinking and existing storage ?
4. Can we shrink? What should be worried about?
5. Are we going to need more Conditions (component conditions) to support this feature ?

PR #12, opened 2021-12-03
Proposal for Quota Reporting and Management

Sunandadadi:


enhancements/quota-management.md
```diff
@@ -0,0 +1,69 @@
...
+### Goals
+
+* A Quay superuser can see the overall storage consumption of the entire registry i.e., an overview of all organizations depicting the aggregate storage consumption of images in an organization.
+* A Quay user can define soft quotas (storage limit) on an organization that will trigger a warning when the set limit is reached.
+* A Quay user can configure alerting thresholds on the storage quota that will allow Quay to send notifications when set threshold is reached.
+* A Quay user can define hard quotas (storage limit) on an organization which will not allow them to push any more images when set threshold is reached.
+
+### Non-Goals
+
+??
```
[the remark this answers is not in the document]
Sunandadadi: Thank you. I've added this here.

enhancements/quota-management.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Motivation
+
+Users can create new organizations and repositories in their home organizations without any practical storage limitation.
+On-prem Quay customers are subject to capacity limits of their environment, specifically backend storage consumption.
+No knowledge about their storage consumption makes it hard for a service owner of a Quay registry to define SLAs and maintain a budget.
+
+### Goals
+
+* A Quay superuser can see the overall storage consumption of the entire registry i.e., an overview of all organizations depicting the aggregate storage consumption of images in an organization.
```
kleesc, line 54: A user should also have access to their own resource usage via the API.
So a user should be able to query `/api/v1/repository/<namespace>/<reponame>/usage` or something similar to get the status of his repo.

kleesc: The superuser, or ui view, would do some sort of aggregation on these.

enhancements/quota-management.md
```diff
@@ -0,0 +1,103 @@
...
+## Design Details
+
+Initial work is to add Quota Reporting on organizations i.e., providing Quay users with their storage consumption data on a Quay organization.
+In the next steps, we will add Quota Enforcements i.e., soft and hard checks that notify or restrict Quay users from pushing images to registry.
+
+The expected design flow is depicted as below:
+![](https://user-images.githubusercontent.com/11522230/144654849-62d00b53-3be7-4009-afb6-2523f2931841.jpeg)
+NOTE: The boxes with the black border show the current flow, and the boxes with the green border show what needs to be implemented to support this feature. Design credits: @kwestpha
+
+The table RepositorySize (to be created) stores the storage consumption(in KBs) of a Quay repository within an organization. The sum of all repository sizes for an organization,
```
kleesc, line 73: We should keep track of sizes in bytes, just for consistency, since that's the unit we use elsewhere in Quay. Any conversion if needed is easy enough to do.  A bigint should be more than enough to cover any reasonable use cases.

enhancements/quota-management.md
```diff
@@ -0,0 +1,103 @@
...
+defines the current storage size of a Quay organization.
+When an image push is initialized, the user's organization storage is validated to check if it is beyond the configured Quota limits.
+If yes, for a soft check, user is notified and, for a hard check, the push is stopped. If storage consumption is within configured Quota limits,
+the push happens as expected.
+
+Images are uploaded to the Quay registry as chunks. Each of these chunk upload triggers, the next flow.
+The size of each image chunk is stored a temporary table, TemporaryUsage (to be created) that has a foreign key reference to the earlier mentioned RepositorySize table.
```
kleesc, line 80: Do we need to track this in a separate table? BlobUpload already tracks the size of a blob upload.

enhancements/quota-management.md
```diff
@@ -0,0 +1,103 @@
...
+This size of every chunk upload is added to the current repository size that calculates a temporary repository size. This temporary size is validated to be
+within the configured Quota limits. If this temporary size reaches beyond the configured Quota limit, an exception is raised.
+
+Images manifest deletion follows the same flow as on Quay where the link between associated tags and the manifest are deleted. One addition to this,
+is, after the manifest is deleted, the repository size is recalculated and updated in the RepositorySize table.
+This is depicted below:
+![](https://user-images.githubusercontent.com/11522230/145113938-7a8e3fcf-693a-488e-a711-bcf116e04446.png)
+
+### Risks and Mitigations
```
kleesc, line 89: A few caveats with doing the calculations on push vs a background worker: existing images will not have any usage data. This means that a backfill for all existing repositories will be required _before_ the feature is enabled. This also means that this calculation becomes part of a push's critical path, otherwise the usage data could possibly drift.

kleesc: We should try and define how to reconcile different states of a Quay installation into one where quota reporting can be enabled. For example, what happens if a user enables quota reporting, disables it and reanables it, etc, ...

enhancements/quota-management.md
Sunandadadi: Yes, agreed. Will add this in goals.

enhancements/quota-management.md
kwestpharedhat: The data required exists on all old data already. This new path does not become critical as the data comes from elsewhere the reporting can fall back upon slower methods of reporting if it becomes necessary

enhancements/quota-management.md
Sunandadadi: Didn't know about the BlobUpload table. That will work. Will update the design accordingly.

enhancements/quota-management.md
Sunandadadi: Regarding enabling and disabling Quota: I am thinking in the terms of having a button on an organization say `Sync Quota` when Quota is enabled. When clicked, we calculate the Quota usage for the org and show. If user disables, Quota is not calculated after disabling, but when re-enabled again, they have the option to sync it back. 
But, if hard limits are configured and user disables Quota, I am assuming that the limits should not hold true and the user can push images.

PR #13, opened 2021-12-10
Proposal for Quay as a Proxy Cache

Sunandadadi:


enhancements/proxy-cache.md
```diff
@@ -0,0 +1,69 @@
...
+* A proxy cache organization will transparently cache and stream images to client. The images in the proxy cache organization should
+  support the default expected behaviour (like security scanner, time machine, etc) as other images on Quay.
+* Given the sensitive nature of accessing potentially untrusted upstream registry all cache pulls needs to be logged (audit log).
+* A Quay user can flush the cache to eliminate excess storage consumption.
+* Robot accounts can be created in the cache organization as usual, their RBAC can be managed for all existing cached repositories at a given point in time.
+* Provide metrics to track cache activity and efficiency (hit rate, size, evictions).
+
+### Non-Goals
+
+* In the first phase, configuring a cache proxy organization, caching upstreaming images, and quota management on cached repositories is the target.
```
flavianmissi, line 58: We probably want to add as non-goals push support (cached images should be read-only) and support for [tag expiration](https://docs.projectquay.io/use_quay.html#tag-expiration).
Wdyt?

Sunandadadi: Good point, I agree. I will add that in.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,69 @@
...
+  Other goals will be implemented subsequently based on the work of this proposal.
+
+## Design Details
+
+The expected pull overview is depicted as below:
+![](https://user-images.githubusercontent.com/11522230/145638459-9f7a585c-d6a3-40c5-983f-f4d3a2e90a69.png)
+
+When a user pulls an image from an upstream repository, before pulling, the image is verified against the cached storage.
+If the image does not exist in cached storage, a new image is pulled from the upstream registry and is saved in the cache.
+On the other hand, if the same image exists in cache, the image in cache is verified against the image in the upstream. If both the images are the same,
```
HammerMeetNail, line 68: How will the image be verified against the upstream registry? Will the check against the upstream registry count against an upstream rate limit?

Sunandadadi: That is a good question. Verifying image against the upstream registry can be done by checking the image digest. For the other question, I'd need to check which API calls account towards rate limiting. But, putting this under open questions for now.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,123 @@
+---
+title: Quota Management and Enforcement on Quay
```
flavianmissi, line 2: Title needs updating(:

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,123 @@
...
+  * If the pull happens with the configured staleness period, the image stored in cache is served.
+  * If the pull happens after the configured staleness period, the error is propagated to the user.
+  * This is depicted as below:
+    ![](https://user-images.githubusercontent.com/11522230/145878373-c23d094b-709d-4859-b875-013ea33e34f7.png)
+    Design credits: @fmissi
+
+A quay admin can leverage quota of an organization to limit cache size so the backend storage consumption remains predictable
+by discarding images from the cache according to least recently used frequency.
+This is depicted as below:
+![](https://user-images.githubusercontent.com/11522230/145884935-df19297f-96b5-4c1c-9cdc-e199e04df176.png)
```
flavianmissi, line 99: It'll probably feel weird, but consider crediting yourself as well for the flow with quota, you can think of it as keeping it consistent.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,123 @@
...
+
+* The cached images should have all properties that images on a Quay repository would have.
+* When an upstream image exists in cache, only the updated layers are to be pulled.
+
+### Test Plan
+
+* Implement image pull and save layers to cache.
+* Implement image pull of an outdated image and save only changed layers to cache.
+* Implement configuring of a proxy cache organization.
+* Implement quota management.
```
flavianmissi, line 119: I think the word "implement" is a bit odd in a test plan. What do you think of phrasing it more like "pull a fresh image and check that layers were saved to cache" ? We could also change the section title to something like "Implementation Plan" 🤔

flavianmissi (commented):
This looks great, thanks for taking the time to putting it down into words.
Left some small comments. I'll hold the approval until we discuss it with the rest of the team.
Great work! 🚀

enhancements/proxy-cache.md
Sunandadadi: Oops! changing this.

enhancements/proxy-cache.md
Sunandadadi: Sure, I will put my name there.

enhancements/proxy-cache.md
Sunandadadi: Sure, sounds fine to me.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+Dependencies on container images has increased tremendously with the adoption of container driven development. With the introduction of rate limits
+on popular container registries, Quay will act as a proxy cache to circumvent pull rate limitation from upstream registries. Adding a cache will also
+accelerate pull performance as images are pulled from cache rather than upstream dependencies. Cached images are updated only when the upstream
+image digest differs from cached image. 
+
+### Goals
+
+* A Quay user can define and configure(credentials, staleness period) via config file/app, a repository in Quay that acts a cache for a specific upstream registry.
+* A Quay super can user leverage storage quota of an organization to limit cache size. This means when cache size reaches its quota limit,
```
HammerMeetNail, line 47: nit: this reads a little weird. Is it supposed to be `A Quay superuser can leverage`?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+## Summary
+
+Dependencies on container images has increased tremendously with the adoption of container driven development. With the introduction of rate limits
+on popular container registries, Quay will act as a proxy cache to circumvent pull rate limitation from upstream registries. Adding a cache will also
+accelerate pull performance as images are pulled from cache rather than upstream dependencies. Cached images are updated only when the upstream
+image digest differs from cached image. 
+
+### Goals
+
+* A Quay user can define and configure(credentials, staleness period) via config file/app, a repository in Quay that acts a cache for a specific upstream registry.
```
HammerMeetNail, line 46: Is this saying a Quay repo is mapped to an entire upstream registry, like docker.io? A couple bullets down it sounds like a Quay org is mapped to a registry.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+* Given the sensitive nature of accessing potentially untrusted upstream registry all cache pulls needs to be logged (audit log).
+* A Quay user can flush the cache to eliminate excess storage consumption.
+* Robot accounts can be created in the cache organization as usual, their RBAC can be managed for all existing cached repositories at a given point in time.
+* Provide metrics to track cache activity and efficiency (hit rate, size, evictions).
+
+### Non-Goals
+
+* In the first phase, configuring a cache proxy organization, caching upstreaming images, and quota management on cached repositories is the target.
+  Other goals will be implemented subsequently based on the work of this proposal.
+* Cached images are read-only.
```
HammerMeetNail, line 60: Can you expand on this a little bit with an example?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Design Details
+
+The expected pull overview is depicted as below:
+![](https://user-images.githubusercontent.com/11522230/145866763-58f44c94-839b-4edb-a95b-b9c3648cf187.png)
+Design credits: @fmissi
+
+A user initiates a pulls of an image(say a `postgres:14` image) from an upstream repository on Quay. The repository is checked to see if the image is present.
+1. If the image does not exist, a fresh pull is initiated.
+  * The user is authenticated into the upstream registry and all the layers of `postgres:14` are pulled.
```
HammerMeetNail, line 70: Will upstream auth be required?

HammerMeetNail:
@Sunandadadi This looks good, a few comments. I'm curious, are all of these changes backend? Is there any UI work expected?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+2. If the image exists in cache:
+  * A user can rely on Quay's cache to stay coherent with the upstream source so that I transparently get newer images from the cache
+     when tags have been overwritten in the upstream registry immediately or after a certain period of time.
+  * If the upstream image and cached version of the image are same:
+    * No layers are pulled from the upstream repository and the cached image is served to the user.
+
+  * If the upstream image and cached version of the image are different:
+    * The user is authenticated into the upstream registry and only the changed layers of `postgres:14` are pulled.
+    * The new layers are updated in cache and served to the user in parallel.
```
dmage, line 84: What will happen if the same uncached blob is requested by multiple clients at the same time?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+## Design Details
+
+The expected pull overview is depicted as below:
+![](https://user-images.githubusercontent.com/11522230/145866763-58f44c94-839b-4edb-a95b-b9c3648cf187.png)
+Design credits: @fmissi
+
+A user initiates a pulls of an image(say a `postgres:14` image) from an upstream repository on Quay. The repository is checked to see if the image is present.
+1. If the image does not exist, a fresh pull is initiated.
+  * The user is authenticated into the upstream registry and all the layers of `postgres:14` are pulled.
+  * The pulled layers are saved to cache and served to the user in parallel.
```
dmage, line 71: Will serving and caching reuse the same connection to the upstream registry or will they open two different connections?

enhancements/proxy-cache.md
dmage, line 46: Can you describe how the configuration process will look like? Will it be a parameter in the configuration file or will it be configured in the web console (or Quay API)? If the web console, will it be available for all users or only to the superuser?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+3. If user initiates a pull when the upstream registry which is down:
+  * If the pull happens with the configured staleness period, the image stored in cache is served.
+  * If the pull happens after the configured staleness period, the error is propagated to the user.
+  * This is depicted as below:
+    ![](https://user-images.githubusercontent.com/11522230/145878373-c23d094b-709d-4859-b875-013ea33e34f7.png)
+    Design credits: @fmissi
+
+A quay admin can leverage quota of an organization to limit cache size so the backend storage consumption remains predictable
+by discarding images from the cache according to least recently used frequency.
```
dmage, line 97: If the cache size is set to a very small number (1 MB), will Quay work as a proxy without caching or will it fail?

dmage: Oh, I see, it'll overflow beyond quota limit.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+This is depicted as below:
+![](https://user-images.githubusercontent.com/11522230/145884935-df19297f-96b5-4c1c-9cdc-e199e04df176.png)
+Design credits: @sdadi
+
+A user initiates a pulls of an image(say a `postgres:14` image) from an upstream repository on Quay. If the storage consumption of the organization
+is beyond the quota limit, images in the namespace are removed based on least recently used, to make space for `postgres:14` to be cached.
```
dmage, line 103: Does the current implementation keep track of when the image was last used?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Constraints
+
+* If quota management is enabled with proxy cache organization, and say an org is set to a max quota of 500mb and the image the user wants to pull is 700mb.
+  In such a case, the image pulled will be cached and will overflow beyond quota limit.
+
+### Risks and Mitigations
+
+* The cached images should have all properties that images on a Quay repository would have.
+* When an upstream image exists in cache, only the updated layers are to be pulled.
```
dmage, line 113: Images are immutable. Do you want to say "When the client wants to pull a new image, Quay should reuse already cached layers and download from the upstream repository only new ones."?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+    ![](https://user-images.githubusercontent.com/11522230/145871778-da01585a-7b1b-4c98-903f-809c214578da.png)
+    Design credits: @fmissi
+
+2. If the image exists in cache:
+  * A user can rely on Quay's cache to stay coherent with the upstream source so that I transparently get newer images from the cache
+     when tags have been overwritten in the upstream registry immediately or after a certain period of time.
+  * If the upstream image and cached version of the image are same:
+    * No layers are pulled from the upstream repository and the cached image is served to the user.
+
+  * If the upstream image and cached version of the image are different:
```
dmage, line 82: If the upstream image is deleted? Needs new credentials?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+A user initiates a pulls of an image(say a `postgres:14` image) from an upstream repository on Quay. The repository is checked to see if the image is present.
+1. If the image does not exist, a fresh pull is initiated.
+  * The user is authenticated into the upstream registry and all the layers of `postgres:14` are pulled.
+  * The pulled layers are saved to cache and served to the user in parallel.
+  * This is depicted as below:
+    ![](https://user-images.githubusercontent.com/11522230/145871778-da01585a-7b1b-4c98-903f-809c214578da.png)
+    Design credits: @fmissi
+
+2. If the image exists in cache:
```
dmage, line 76: If the image is pulled by SHA, do we need to perform upstream checks?

enhancements/proxy-cache.md
dmage: It means `Images cannot be pushed into Proxy Cache Organizations`, right?

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+accelerate pull performance as images are pulled from cache rather than upstream dependencies. Cached images are updated only when the upstream
+image digest differs from cached image. 
+
+### Goals
+
+* A Quay user can define and configure(credentials, staleness period) via config file/app, a repository in Quay that acts a cache for a specific upstream registry.
+* A Quay super can user leverage storage quota of an organization to limit cache size. This means when cache size reaches its quota limit,
+  images from cache are evicted based on LRU.
+* A proxy cache organization will transparently cache and stream images to client. The images in the proxy cache organization should
+  support the default expected behaviour (like security scanner, time machine, etc) as other images on Quay.
```
kleesc, line 50: I would expect the time machine policies to be a bit different from the default behaviour, especially since it is expected that there should be some sort of eviction policy on the content pulled from the proxied registries.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+* Given the sensitive nature of accessing potentially untrusted upstream registry all cache pulls needs to be logged (audit log).
```
kleesc, line 51: Pulls are already logged. This can probably be inferred based on the namespace and whether it's set up as a proxy cache or not. Or maybe add additional metadata to the existing pull audit log for these cases.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+* A Quay user can flush the cache to eliminate excess storage consumption.
```
kleesc, line 52: I don't think this should be handled manually by the user. It should be done automatically, based on some policy that we decide on. e.g Removing the oldest cached tags/repos.

enhancements/proxy-cache.md
kleesc: I would assume so, the same way a user would have to provide credentials when mirroring a repo.

enhancements/proxy-cache.md
kleesc: Shouldn't need to. Since digests are content addressable.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+    Design credits: @fmissi
+
+3. If user initiates a pull when the upstream registry which is down:
+  * If the pull happens with the configured staleness period, the image stored in cache is served.
+  * If the pull happens after the configured staleness period, the error is propagated to the user.
+  * This is depicted as below:
+    ![](https://user-images.githubusercontent.com/11522230/145878373-c23d094b-709d-4859-b875-013ea33e34f7.png)
+    Design credits: @fmissi
+
+A quay admin can leverage quota of an organization to limit cache size so the backend storage consumption remains predictable
```
kleesc, line 96: It's not so much quota as it is a "configurable size limit".
Quota imply some sort of limitation on what one can do once a limit is reached.
In this case, it just happens that it's going to use the reporting data from the quota work.

enhancements/proxy-cache.md
kleesc: We log it in the audit logs. But there isn't an easy way to query when a tag was last pulled. And that seems expensive to track. Given that every pull would need to make some sort of update on the database.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+This is depicted as below:
+![](https://user-images.githubusercontent.com/11522230/145884935-df19297f-96b5-4c1c-9cdc-e199e04df176.png)
+Design credits: @sdadi
+
+A user initiates a pulls of an image(say a `postgres:14` image) from an upstream repository on Quay. If the storage consumption of the organization
+is beyond the quota limit, images in the namespace are removed based on least recently used, to make space for `postgres:14` to be cached.
+
+### Constraints
+
+* If quota management is enabled with proxy cache organization, and say an org is set to a max quota of 500mb and the image the user wants to pull is 700mb.
```
kleesc, line 107: Same comment as above as to what quota means in this context.

enhancements/proxy-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+* The cached images should have all properties that images on a Quay repository would have.
+* When an upstream image exists in cache, only the updated layers are to be pulled.
+
+### Implementation Plan
+
+* Pull a fresh image and check that layers are saved to cache.
+* Pull an outdated image and check that only changed layers are saved to cache.
+* Implement configuring of a proxy cache organization.
+* Implement quota management.
```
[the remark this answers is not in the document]
kleesc: Same comment as above as to what "quota" means in this context.

enhancements/proxy-cache.md
flavianmissi: @kleesc do you mean we currently log pulls done by clients pulling from quay?

enhancements/proxy-cache.md
flavianmissi: The automatic removal will exist along side with the manual flush. Manual flush is specifically mentioned in the user stories on the [epic](https://issues.redhat.com/browse/PROJQUAY-465):
> * As a Quay admin I want to be able to leverage the storage quota of an organization to limit the cache size so that backend storage consumption remains predictable by discarding images from the cache according to least recently used or pull frequency
> * [...]
> * As a user I want to be able to flush the cache so that I can eliminate excess storage consumption
If you think this shouldn't be the case then we need to bring it up with Daniel when he's back.

enhancements/proxy-cache.md
flavianmissi: Agreed. This has already been a source of confusion in our last engineering sync.

enhancements/proxy-cache.md
flavianmissi: From the [epic](https://issues.redhat.com/browse/PROJQUAY-465)'s user stories:
> As a administrator I want to be able to select from caching an entire upstream registry (e.g. cache all of docker.io, i.e. docker.io/library/postgres:latest -> quay.corp/cache/library/postgres:latest) or just a selected namespaces (e.g. just cache docker.io/library, i.e. docker.io/library/postgres:latest -> quay.corp/docker-cache/postgres:latest) so that I can constrain access to potentially untrusted upstream registries

Maybe we should add all users stories to the enhancement so there's no confusion.
Wdyt?

flavianmissi: Good point, we need to at least mention how the setup will be addressed here.
From the user stories in the [epic](https://issues.redhat.com/browse/PROJQUAY-465), I understand that regular users should be able to set up proxy orgs:
> * As a Quay user I want to be able to define an organization in Quay that acts a cache for a specific upstream registry
> * As a Quay user I want to be able to supply credentials to the upstream registry when defining a cache organization so that I can circumvent / extend possible pull-rate limits or access private repositories

enhancements/proxy-cache.md
flavianmissi: I think time machine capabilities for the cache was something that Daniel mentioned as being desirable. It does seem like it would conflict, or at least be a source of confusion, with regards to the staleness period though.
@Sunandadadi maybe we mention this in the open questions?

enhancements/proxy-cache.md
flavianmissi: Yup, what @dmage said.

enhancements/proxy-cache.md
flavianmissi: It shouldn't be a requirement though, docker hub for example supports anonymous pulls (although you'd still need a token, it'd just be an anonymous one). Autheticantion with the upstream registry should be required when users want to cache private repostories.

enhancements/proxy-cache.md
flavianmissi: I have not looked in depth at the technical feasibility of this, but my idea is to pull a layer from the upstream registry and then cache and stream it. So there should only be one pull against the upstream registry and we'd use the result of that to both cache and stream the layer to the user.

enhancements/proxy-cache.md
flavianmissi: Depending on the staleness period the client would either get the error propagated to them (image outside staleness period) or get the version quay has in cache (image within staleness period).

enhancements/proxy-cache.md
flavianmissi: I'm not sure. I suspect that the desired behaviour is that we don't end up with duplicates blobs. Do we care about how we end up with that result? 🤔

enhancements/proxy-cache.md
flavianmissi: We're kind of dependant on that information though, we could also use pull frequence but I'm not sure we have that either.
@Sunandadadi I think we need to add this one as an open question as well.

enhancements/proxy-cache.md
flavianmissi: I like @dmage's suggested rephrasing, let's use that instead @Sunandadadi ?

enhancements/proxy-cache.md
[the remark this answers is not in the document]
flavianmissi: We do depend on quota reporting, since that will introduce the repository size calculations. The pull thry proxy [epic](https://issues.redhat.com/browse/PROJQUAY-465) is currently "blocked by" the quota epic. It doesn't hurt to mention it here though.

enhancements/proxy-cache.md
dmage: I think it should be the other way around: first we need an enhancement proposal that describes a feature the way it will look in the next release, then we discuss it and accept it, and then we can create stories and implement it.

The enhancement proposals should be detailed enough so that they are useful to the documentation team, QA, support engineers, early adopters that use nightly builds. If we change some of our decisions, we need to update the EPs to keep them useful.

Note: some my comments have questions that are beyond the detalisation of EPs. I'm trying to understand it deeper and find problems that can make it unimplementable. Implementation details that are not visible to users can be omitted in EPs, but we still may need to discuss them.

enhancements/proxy-cache.md
Sunandadadi: Yes thanks for pointing it out, looks like a typo. I will change this.

enhancements/proxy-cache.md
Sunandadadi: I've added what @dmage mentioned, for better clarity.

enhancements/proxy-cache.md
Sunandadadi: Yes, agreed, I will add what you've mentioned here @flavianmissi.

enhancements/proxy-cache.md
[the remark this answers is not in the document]
Sunandadadi: I've added the dependency here.

enhancements/proxy-cache.md
Sunandadadi: Good point. I will look into this and update once we have a clearer understanding.

enhancements/proxy-cache.md
Sunandadadi: Noted, I will add it.

enhancements/proxy-cache.md
Sunandadadi: Agreed, thanks for that @dmage

enhancements/proxy-cache.md
Sunandadadi: Yes, agreed, adding it.

enhancements/proxy-cache.md
flavianmissi: I understand this as being about logging the pulls performed by quay against the upstream registry, not necessarily the pulls performed by the client.

enhancements/proxy-cache.md
Sunandadadi: Agreed, I will change quota to `configurable size limit`

flavianmissi (approved):
Great work, Sunanda! 🎉
Let's iterate on this as necessary with the rest of the team after holidays.

Sunandadadi:
> @Sunandadadi This looks good, a few comments. I'm curious, are all of these changes backend? Is there any UI work expected?

@HammerMeetNail Yes, there will be UI work to configure a proxy cache organization where users can configure credentials, upstream repo path, staleness period and such

HammerMeetNail (approved):
LGTM 👍 

Can we back burner any UI work that's needed? There's a frontend rewrite planned for next year that this work might fall nicely into.

PR #14, opened 2021-12-21
Proposal to Add Build Option Without Bare-metal Constraint

bcaton85:
Proposal of solution to remove bare-metal constraint from builds.

### DIFF
--- a/enhancements/builders.md
+++ b/enhancements/builders.md
@@ -1,0 +1,84 @@
+---
+title: Proposal to Add Build Option Without Bare-metal Constraint
+authors:
+  - "@bcaton"
+reviewers:
+  - "@kleesc"
+  - "@bcaton"
+  - "@harishsurf"
+approvers:
+  - "@kleesc"
+  - "@bcaton"
+  - "@harishsurf"
+creation-date: 2021-12-03
+last-updated: 2022-01-06
+status: implementable
+---
+
+# remove-bare-metal-build-constraint
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Open Questions
+
+> 1. Will we encounter issues down the line running builds in unpriviliged containers with Podman?
+> 1. Does running builds in unprivileged containers give us the build isolation we need?
+
+## Summary
+
+Currently Quay builds run Podman commands in virtual machines launched via Pods. To run builds on virtual platforms this would require enabling nested virtualization, which is not possible in RHEL or OCP. This places the requirement of builds having to run on bare-metal clusters, which are a precious resource. This proposal is to remove the bare-metal constraint required to run builds by adding an additional build option which does not contain the virtual machine layer.
+
+### Goals
+
+- Builds are able to run on virtualized platforms
+- Have backwards-compability to run previous build configurations
+
+## Design Details
+
+This solution proposes adding an additional build option to remove the VM layer to permit builds on virtualized platforms. This will take the form of an additional executor added to the build manager. An unprivileged Podman container will be used to run the image build.
+
+This diagram describes the current workflow:
+
+![image](https://user-images.githubusercontent.com/22058392/146998028-60e79958-5ef8-49ea-af07-7c9aaeef938a.png)
+
+The following diagram describes the proposed workflow:
+
+![image](https://user-images.githubusercontent.com/22058392/146998050-32168e74-dbdf-4193-aa10-c28ac52f182a.png)
+
+The changes to the workflow can be described as:
+1. The build manager will create the job object.
+2. The job object will create a pod using the **quay-builder image**.
+    - The **quay-builder image** will contain the **quay-builder binary** and the **Podman service**.
+    - This pod will run as unprivileged
+3. The **quay-builder binary** will run commands to build the image while also communicating status and retrieving build info from the build manager
+
+A quick POC has been completed where a very basic Dockerfile was submitted through the UI and the build was completed in a virtual environment in an unprivileged container. 
+
+- [Dockerfile producing the **quay-builder image**.](https://github.com/bcaton85/quay-builder/blob/podman-build/Dockerfile.podman.secure)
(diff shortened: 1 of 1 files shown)


enhancements/builders.md
```diff
@@ -0,0 +1,82 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Open Questions
+
+> 1. Will we encounter issues down the line running builds in unpriviliged containers with Podman?
+> 1. Does running builds in unprivileged containers give us the build isolation we need?
+
+## Summary
+
+Currently Quay builds run Podman commands in virtual machines launched via Pods. To run builds on virtual platforms this would require enabling nested virtualization, which is not possible in RHEL or OCP. This places the requirement of builds having to run on bare-metal clusters, which are a precious resource. This proposol is to remove the bare-metal constraint required to run builds.
```
HammerMeetNail, line 34: Mind calling out that this proposal is for an additional method of building images? 

> remove bare-metal constraint
^ This wording makes me wonder if bare-metal builds are going away.

enhancements/builders.md
```diff
@@ -0,0 +1,82 @@
...
+![image](https://user-images.githubusercontent.com/22058392/146998050-32168e74-dbdf-4193-aa10-c28ac52f182a.png)
+
+The changes to the workflow can be described as:
+1. The build manager will create the job object.
+2. The job object will create a pod using the **quay-builder image**.
+    - The **quay-builder image** will contain the **quay-builder binary** and the **Podman service**.
+    - This pod will run as unprivileged
+3. The **quay-builder binary** will run commands to build the image while also communicating status and retrieving build info from the build manager
+
+A quick POC has been completed where a very basic Dockerfile was submitted through the UI and the build was completed in a virtual environment in an unprivileged container.
```
HammerMeetNail, line 60: Will any additional UI work be needed?

enhancements/builders.md
```diff
@@ -0,0 +1,82 @@
...
+- [Dockerfile producing the **quay-builder image**.](https://github.com/bcaton85/quay-builder/blob/podman-build/Dockerfile.podman.secure)
+- [Changes made to Quay to run the new **quay-builder image**.](https://github.com/quay/quay/pull/1034/files)
+- Test Dockerfile used:
+  ```Dockerfile
+  FROM hello-world 
+
+  ENV TEST_VAR=test
+  ```
+
+### Constraints
```
HammerMeetNail, line 71: Are there any additional privileges that need to be granted in order to make this work or is implementation as easy as stripping out the existing VM layer?

enhancements/builders.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Design Details
+
+This solution proposes removing the VM layer to permit builds on virtualized platforms. An unprivileged Podman container will be used to run the image build.
+
+This diagram describes the current workflow:
+
+![image](https://user-images.githubusercontent.com/22058392/146998028-60e79958-5ef8-49ea-af07-7c9aaeef938a.png)
+
+The following diagram describes the proposed workflow:
```
HammerMeetNail, line 49: What are the benefits of doing this in house vs leveraging something like Tekton? Faster and easier to deliver?

enhancements/builders.md
bcaton85: No UI work will be needed, only requires changes to the Quay configuration.

enhancements/builders.md
bcaton85: Removing the VM constraint requires straight forward code changes. Integrating Tekton would require overhauling the current build process. Tekton solves the issue of us handling the current build workflow, but is not necessary for removing the build constraint. We're currently seeing these as two different solutions that are solving two separate problems.

enhancements/builders.md
bcaton85: the `anyuid` policy needs to be added to the service account running the builds. [Docs here.](https://github.com/containers/buildah/blob/main/docs/tutorials/05-openshift-rootless-build.md#create-service-account-for-building-images) Here's the command:
`oc adm policy add-scc-to-user anyuid -z buildah-sa`
This would have to be done if we were to use Tekton as well. Other than that it's just pointing quay to use a different builder image.

enhancements/builders.md
bcaton85: Updated.

enhancements/builders.md
HammerMeetNail: If it's an additional method for running builds, won't some UI work be needed? Otherwise how will users pick the new build method?

bcaton85: as of now the builds are configured entirely through the quay configuration. Now that you mention it there may need to be changes to the config tool, but not to quay UI itself

bcaton85: Yes, it would be a separate executor

bcaton85: @kleesc can keep me honest

enhancements/builders.md
```diff
@@ -0,0 +1,83 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Open Questions
+
+> 1. Will we encounter issues down the line running builds in unpriviliged containers with Podman?
+> 1. Does running builds in unprivileged containers give us the build isolation we need?
+
+## Summary
+
+Currently Quay builds run Podman commands in virtual machines launched via Pods. To run builds on virtual platforms this would require enabling nested virtualization, which is not possible in RHEL or OCP. This places the requirement of builds having to run on bare-metal clusters, which are a precious resource. This proposol is to remove the bare-metal constraint required to run builds by adding an additional build option which does not contain the virtual machine layer.
```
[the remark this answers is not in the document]
bcaton85: fixed

bcaton85:
Running the builds in the virtual machine will still be provided as an option. This will be an additional build option for those who are willing to run builds directly in unprivileged containers.

HammerMeetNail:
> This will be an additional build option for those who are willing to run builds directly in unprivileged containers.

Maybe we should call out that this will inherently be a less secure option and should only be used for trusted code?

bcaton85:
Agreed, i'm assuming the best place for that would be the documentation for this option

HammerMeetNail:
@bcaton85 are we good to merge this?

bcaton85:
yes

PR #15, opened 2022-01-17
Clair Unmanaged Database Proposal

jonathankingfc:


enhancements/unmanaged-clair-database.md
```diff
@@ -0,0 +1,142 @@
...
+| Unmanaged | Managed | Yes | Error, unmanaged Clair should not have a clair config |
+```
+
+This matrix can be tested using unit tests.
+
+In addition to this, e2e tests should be implemented inside a real cluster. This can be done using the `kuttl` tool. A PR has been opened to add the `kuttl` e2e tests to the operator and can be found at https://github.com/quay/quay-operator/tree/kuttl
+
+### Upgrade / Downgrade Strategy
+
+Since previous versions of Quay do not contain a `clair-config.yaml` in the `config-bundle-secret`, and do not have a customizable `clairdatabase` component, upgrades should not affect the behavior of any `QuayRegistry`. After an upgrade, the operator should automatically set `clairdatabase` to be a managed component, and the default generation of the clair config should proceed as normal.
```
ricardomaraschini, line 118: Small consideration here, when you say: "After an upgrade, the operator should automatically set `clairdatabase` to be a managed component" I would add "if Clair is managed". If during the upgrade clair is not managed then we don't need to spawn its database. wdyt?

enhancements/unmanaged-clair-database.md
```diff
@@ -0,0 +1,142 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Open Questions
+
+> 1.  Should the clair config be fully customizable, or should it be limited to specific fields such as the DSN?
```
ricardomaraschini, line 27: I think we should allow any config here. We might choose the ones we want to validate (like the DSN for instance).

enhancements/unmanaged-clair-database.md
```diff
@@ -0,0 +1,142 @@
...
+
+```
+
+The provided clair-config.yaml would be mounted into the Clair pod, and any fields not provided should be automatically populated with defaults using the Clair config module.
+
+### Implementation Details/Notes/Constraints
+
+One implementation detail to note is that a managed Clair component should consume a custom `clair-config.yaml` as long as it is provided. This is similar to how the base Quay component can be customized with a `config.yaml` despite always being a managed component. However, the clair config must be in sync with the state of the `clairdatabase` component. For example, if a user chooses a managed `clairdatabase` component, the `clair-config.yaml` should not contain custom DSN values.
+
+### Risks and Mitigations
```
ricardomaraschini, line 91: There is one extra concern that I haven't seen you mention in this document: do we need to change anything in the config-tool to support this new feature?

jonathankingfc: The new config module comes with some schema validation. We would need to validate the database connection in addition to this. This could be done using the config-tool module that exists today. See this function here -> https://github.com/quay/config-tool/blob/171819fa5bca7b3a55be146766abd453dcab43b1/pkg/lib/shared/validators.go#L458. We will likely need to change it around a bit to be more generic, but the validation logic can be reused.

enhancements/unmanaged-clair-database.md
jonathankingfc: Sounds good. The WIP implementation is currently doing this

enhancements/unmanaged-clair-database.md
jonathankingfc: Good point, I will add this

PR #24, opened 2023-07-24
Quay API v2

Sunandadadi:


enhancements/api-v2.md
```diff
@@ -0,0 +1,122 @@
...
+
+    a. prev_cursor: encrypted string to fetch previous page results
+
+    b. next_cursor: encrypted string to fetch next page results
+
+Response body:
+```
+{
+  results: [items per page],
+  total_count: count of all items on all pages,
```
dmage, line 75: Will Quay count all objects for every request? Does Quay have this information today or will it be a new extra query to db?

enhancements/api-v2.md
```diff
@@ -0,0 +1,122 @@
...
+  `v2_draft` will be replaced with `v2` when ready to release. 
+- Existing decorators like the scope of API, page parameters will be reused where ever required.
+
+### Pagination
+
+It is difficult to say that one pagination strategy suits all needs. After understanding the pros and cons of various pagination strategies, 
+cusor-based pagination seems to work best for our current needs.  
+
+Cursor will contain a unique sequential db column to base cursors on along with additional data when required.
+(Eg: sorting results on non-unique keys, like datetime)
```
dmage, line 61: Please add examples of the cursor data.

I suspect there will be two options:
* `{"id": 123}` where 123 is the id of the last item on the return page;
* `{"id": "2023-08-02T12:34:56Z", "n": 32}` where n is how many items were sent to the client with the same id (on the current page and possible on previous pages too).

But I do not want to speculate on this and I want us all to be on the same page. This detail is really important.

enhancements/api-v2.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Filtering
+
+Typical syntax of an API filter will look like: `key1[operation1]=<value1>&key2[operation2]=<value2>` where:
+- key: is the field on which the search will be performed (Eg: repo name, org name, etc)
+- operation: can be different based on the datatype of the key. (Eg: for integers, `lt|lte|gt|gte`, for strings: `like|eq|regex`)
+- value: is what the query will filter for
+- multiple supported filters can be added to the query parameters using `&`
+
+Eg: (GET `/tags?tag_name[like]=quay`, GET `/logs?created_at[lte]=2023-07-20 19:56`)
```
dmage, line 91: Do we have examples when multiple filters are needed for the UI?

enhancements/api-v2.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Sorting
+
+Sort order will be represented by the query parameter `sort`, ascending order by `+` and descending order by `-`.
+Multiple sorts can be passed in the url by adding comma separated keys.
+Eg: (GET `/tags?sort=+created_at`, GET `/users?sort=+creation_date,-last_accessed`)
```
dmage, line 97: Will users be able to sort by any column? How do we deal with columns that are not indexed and sorting on which can be terribly slow?

enhancements/api-v2.md
Sunandadadi: This will be counting all objects for each request. We can add a caching layer for frequent requests. I don't think we have this in the code today, we'd need make an extra query here.

enhancements/api-v2.md
Sunandadadi: Surely, good point, will add examples.

enhancements/api-v2.md
Sunandadadi: I don't see a use case of multiple filters in the designs. I want the design to accommodate this as this will be a one time investment. Good to have for future needs.

enhancements/api-v2.md
Sunandadadi: My thoughts are to restrict sorts on only the cols the UI needs today. If a user makes a request with unsupported col sort, 400 will be sent back.

Sunandadadi: I will document the above.

enhancements/api-v2.md
```diff
@@ -0,0 +1,140 @@
...
+
+It is difficult to say that one pagination strategy suits all needs. After understanding the pros and cons of various pagination strategies, 
+cusor-based pagination seems to work best for our current needs.  
+
+Cursor will contain a unique sequential db column to base cursors on along with additional data when required.
+For example, sorting results on non-unique keys, like datetime (currently used in the UI for sorting usage logs)
+
+Cursor data is extensible and can look like:
+```
+{"id": 123}
```
dmage, line 66: Can a cursor from the request with one set of query parameters be reused with the request with another set of parameters (with different sorting or different filters)?

enhancements/api-v2.md
```diff
@@ -0,0 +1,140 @@
...
+{"id": "2023-08-02T12:34:56Z", "offset": 20}
+{"id": "2023-08-02T12:34:56Z", "some_other_key": val1, "another_key": val2}
+```
+
+The following will be expected request parameters for an API that is to be paginated:
+1. limit: number of items to be returned for the page, max limit set to 100 items per page (based from current Quay API guidelines).
+2. prev_cursor/next_cursor
+
+    a. prev_cursor: encrypted string to fetch previous page results
```
dmage, line 75: How will this cursor be generated?

enhancements/api-v2.md
```diff
@@ -0,0 +1,140 @@
...
+- Proposed path of the endpoints: In an attempt to not change the existing endpoints, new APIs will be under `/api/v2_draft/`.
+  `v2_draft` will be replaced with `v2` when ready to release. 
+- Existing decorators like the scope of API, page parameters will be reused where ever required.
+
+### Pagination
+
+It is difficult to say that one pagination strategy suits all needs. After understanding the pros and cons of various pagination strategies, 
+cusor-based pagination seems to work best for our current needs.  
+
+Cursor will contain a unique sequential db column to base cursors on along with additional data when required.
```
kleesc, line 61: How is this different from what we currently support in terms of pagination?
On some resources, we already have cursor based pagination, where we return an opaque token to be sent on subsequent requests.

How is this proposal different or supposed to improved what's already there, performance or otherwise?

enhancements/api-v2.md
```diff
@@ -0,0 +1,140 @@
...
+Cursor data is extensible and can look like:
+```
+{"id": 123}
+{"id": "2023-08-02T12:34:56Z", "offset": 20}
+{"id": "2023-08-02T12:34:56Z", "some_other_key": val1, "another_key": val2}
+```
+
+The following will be expected request parameters for an API that is to be paginated:
+1. limit: number of items to be returned for the page, max limit set to 100 items per page (based from current Quay API guidelines).
+2. prev_cursor/next_cursor
```
kleesc, line 73: What's the use case for a pointer to the previous page?

enhancements/api-v2.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Summary
+
+Currently, only a few APIs have the option to paginate results on the backend. The goal of the proposal is to have a re-usable system to perform actions on the backend and send the curated result on the frontend. Ideally frontend should only display the results and allow the backend to perform
+all the heavy lifting.
+
+## Motivation
+
+In the quay ui, filtering, sorting and occasionally paginating the results happen on the frontend. 
+This leads to high rendering time on the frontend and thereby a poor user experience.
```
kleesc, line 39: What are the implication of delegating that work to the server-side?
Can you add some implementation details w.r.t the database (query changes, indices, ...) involved,

enhancements/api-v2.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Goals
+
+Implement a generic system that is consistent across different endpoints to support the following operations on Quay endpoints:
+- pagination
+- filtering
+- sorting
```
kleesc, line 46: See comment above. I assume doing filtering and sorting on the server side vs client-side has some significant performance trade-offs.

enhancements/api-v2.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Design Details
+
+### API Design
+
+- Proposed path of the endpoints: In an attempt to not change the existing endpoints, new APIs will be under `/api/v2_draft/`.
+  `v2_draft` will be replaced with `v2` when ready to release. 
+- Existing decorators like the scope of API, page parameters will be reused where ever required.
```
kleesc, line 54: Is this not going to be backward-compatible? If just adding pagination to the current existing resources.

PR #25, opened 2023-08-15
proposal: Tag Auto-pruning

bcaton85:


enhancements/auto-pruning.md
```diff
@@ -0,0 +1,149 @@
...
+| is_running | boolean | default false | Policy is currently being executed |
+
+### Auto-pruning worker
+
+The auto-pruning worker is an asyncronous job that executes the policy and performs the following steps:
+* Worker starts every 30s or to some configured number
+* Select a policy with the least `last_ran_ms` and `is_running`=`false`
+    * A policy with a null `last_ran_ms` will also be selected first as this indicates it was never ran. This prioritizes policies that haven't been ran recently.
+* Set the policy to `is_running` = `true`
+* Determine which pruning method to use based on `config.pruneBy`
```
dmage, line 127: What if there are two policies defined?

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+* Prune tags older than date, provided the number of days 
+
+For this implementation users will only be able to configure one policy per namespace. To interface with this new UI there will be a new set of CRUD endpoints for managing auto-pruning policies. The endpoints returning lists will be paginated.
+
+**Create policy**
+```
+POST /api/v1/organization/{organization}/autoprunepolicy
+By date:
+{
+    pruneBy: "creation_date",
```
hdonnay, line 71: `PruneBy` should just be `policy` or similar -- consider some future policy that looks at multiple data points.

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+
+**autoprunetask**
+
+This table registers tasks to be executed by the auto-prune worker. Tasks are executed within the context of a single namespace. Only one task per namespace will ever exist.
+
+| Field | Datatype | Attributes | Description |
+| --- | ----------- | ----------- | ----------- |
+| namespace_id | integer | FK, not null | Namespace that this task belongs too |
+| last_ran_ms | bigint | nullable, indexed | Last time the worker executed the policies for this namespace |
+| is_running | boolean | default false | Policies for this namespace are currently being executed |
```
hdonnay, line 125: `Is_running` will incur an extra write on every execution, so I'd suggest against it.
The application should be able to work out if the task is running if the row is locked (e.g. if `SELECT 1 FROM autoprunetask WHERE namespace_id = $1 SKIP LOCKED;` returns nothing).

This table is also missing reporting for the status of the previous run; I'd suggest adding a `TEXT` column for holding an error return (NULL meaning no error), or a `TEXT` and a `BOOLEAN` for "status" and "errored".

enhancements/namespace-auto-pruning.md
bcaton85: yup good point

enhancements/namespace-auto-pruning.md
bcaton85: > The application should be able to work out if the task is running if the row is locked

When exactly does the row get locked in the first place? At the start of the transaction? When exactly does that lock get released?

> This table is also missing reporting for the status of the previous run; I'd suggest adding a TEXT column for holding an error return

Agreed, good idea

hdonnay: The query feeding the worker has some row-level locking, right? The way this is done with PostgresQL is `SELECT ... FROM ... WHERE ... LIMIT ... FOR UPDATE SKIP LOCKED;` (specific strength of lock depends).

bcaton85: ah okay i see now. Yeah that's a much better method. Looks like there's methods already in quay supporting that. I'll use those instead

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Get policy**
+```
+GET /api/v1/organization/{organization}/autoprunepolicy/{policyid}
+```
+
+**Delete policy**
+```
+DELETE /api/v1/organization/{organization}/autoprunepolicy/{policyid}
+```
```
kleesc, line 102: What is the unit of work of these policies, tags or manifests?
For example, if multiple tags of different ages point to a single manifest, then these policy wouldn't do much in terms of reclaiming storage space.

enhancements/namespace-auto-pruning.md
kleesc, line 125: Couldn't this be done as a `queueitem`, and implemented as a queueworker?

enhancements/namespace-auto-pruning.md
bcaton85: Tags. You're correct but I think the tags pointing to the same manifest would be created at similar times meaning they would be pruned by a similar time as well. This would probably be something to document for the user.

enhancements/namespace-auto-pruning.md
bcaton85: After the queue item is processed successfully isn't it deleted? This task needs to be re-ran after some time.

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+## Open Questions [optional]
+
+1. Should there be a dry run feature to test the policy?
+1. Should this only be implemented in the new UI and not the old UI?
+1. Can a user configure multiple policies per namespace?
+1. Can a user specify tags or repositories by pattern?
+
+## Summary
+
+In order for a container registry to run cost effectively unused resources should be deleted to reclaim storage space. This task should be done automatically in Quay by deleting unused tags through a given policy. The policy used should be configured by the user and a variety policy options should be given. Examples of initial policies would be prune tags older than X number of days or prune tags when more than 100 are present.
```
syed, line 34: Apart from a policy configured by the user. There should also be a system wide default policy that the admin should be able to apply

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Motivation
+
+Reclaiming space is essential in running a cost efficient registry. Having this done automatically within Quay removes the need for manual intervention or custom software.
+
+### Goals
+
+* User is able to provide a time based policy for auto-pruning tags within a namespace
```
syed, line 42: User should also be able to specify pruning on a repo basis. I can imagine a case where a namespace has a "dev" repo which may periodically need pruning and a deploy repo which doesn't need pruning

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+* User is able to provide a tag number based policy for auto-pruning tags within a namespace
+* User is able to specify whether tags go into the time machine or are immediately deleted
+* The correct tags are deleted when the policy is applied
+* Pruned tags are tracked via usage logs
+
+### Non-Goals
+
+* Allowing users to specify multiple policies per namespace
+* Applying policies based on repo or tag name pattern matching
+* Allowing policy configurations at the repository level
```
syed, line 52: why not at repo level?

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Managing Policies
+
+A user is able to configure auto-pruning within the namespace settings. To start, there will be two auto-pruning options:
+* Prune by number of tags, provided the number of tags
+* Prune tags older than date, provided the number of days 
+
+For this implementation users will only be able to configure one policy per namespace. To interface with this new UI there will be a new set of CRUD endpoints for managing auto-pruning policies. The endpoints returning lists will be paginated.
+
+**Create policy**
```
syed, line 66: To make it generic, you can define two options in your API one is the `policy` and the other one is `action`. A `policy` can be something like you described (age > X or count > Y) and the `action` could be either "delete immediately" or "put in time machine for x days" or it could also be something else

enhancements/namespace-auto-pruning.md
syed: Quay uses redis for locks in parts of the code like GC for example

enhancements/auto-pruning.md
bcaton85: multiple policies are supported in the redesign

enhancements/namespace-auto-pruning.md
bcaton85: it looks like the queue worker uses the `for update` @hdonnay is describing. Would we want to use that instead to remove the dependency on redis?

enhancements/namespace-auto-pruning.md
bcaton85: Yup that's right. That's something that will be implemented in the future.

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+
+**namespaceautoprunepolicy**
+
+This table holds the policy configuration for a single namespace. For this proposal there will be only one entry per namespace but there will be support for multiple rows per namespace_id. The config field will hold the policy details such as `{pruneBy: "creation_date", olderThan: "2w"}` or `{pruneBy: "number_of_tags",  numTags: 100}`.
+
+| Field | Datatype | Attributes | Description|
+| --- | ----------- | ----------- | ----------- |
+| uuid | character varying(255) | Unique, not null, indexed | Unique identifier for this policy |
+| namespace_id | integer | FK, not null | Namespace that the policy falls under |
+| config | text | JSON, not null | Policy configuration |
```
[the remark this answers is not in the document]
bcaton85: agreed

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+
+**autoprunetask**
```
syed, line 117: maybe `autoprunetaskstatus` like we have `manifestsecuritystatus`

enhancements/namespace-auto-pruning.md
Sunandadadi, line 125: In the future when we start supporting multiple policies for a namespace, since neither of the cols are not unique, there can be a possibility of collisions. 
Can we have a FK, reference to `namespaceautoprunepolicy.uuid` in `autoprunetask`? So, we know which policy is running & can be useful when debugging too as to why a policy is stuck/did not run.

enhancements/namespace-auto-pruning.md
Sunandadadi, line 117: Do we want to add a `priority` field here? The priority will determine the order in which items are picked from the table. Eg: paying customers can have p1 while free customers p2. Just something to think about.

bcaton85: Interesting idea. I think that's something worth investigating but is most likely out of scope for now. As of now priority goes to tasks that haven't been ran or haven't been ran in the longest amount of time.

enhancements/namespace-auto-pruning.md
bcaton85: My only concern with this is it makes the endpoints more complicated. I would say the policy is the method (age, count) + action + extra parameters all together. Similar to a mirror config rule.

enhancements/namespace-auto-pruning.md
bcaton85: The rows in this table are for running policies for a specific namespace and not a specific policy, so there would be no duplicate `namespace_id` values. When we support multiple namespace policies we'll fetch them all in [this step](https://github.com/quay/enhancements/pull/25/files#diff-355ab33b15f7b6f2ac5e26ca752cbd9d70fc3d9f60d03d84f70f55f47cf32c53R145) instead of just the one. We can't have a FK for a task because a task will eventually run multiple policies at once. For debugging info i'll add a status field as hank suggested that will have details if there's an error.

Sunandadadi: Ah! ok, sounds good, I got it wrong.

enhancements/namespace-auto-pruning.md
Sunandadadi: ok, may be we can add this under open questions?

enhancements/namespace-auto-pruning.md
bcaton85: No problem!

enhancements/namespace-auto-pruning.md
bcaton85: Yes, I'll add it there

bcaton85: renamed

enhancements/namespace-auto-pruning.md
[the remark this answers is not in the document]
bcaton85: renamed

enhancements/namespace-auto-pruning.md
bcaton85: renamed to method

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,172 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Open Questions [optional]
+
+1. Should there be a dry run feature to test the policy?
+1. Should this only be implemented in the new UI and not the old UI?
+1. Can a user configure multiple policies per namespace?
+1. Can a user specify tags or repositories by pattern?
+1. Do we want the ability to have tiered priority for auto-prune execution?
+
```
BillDett, line 32: How do these autopruning policies work in a Proxy Organization? Proxy Organizations already have a form of autopruning in effect if storage quota is enabled on that Org. If we allow these autopruning policies on a Proxy Organization with quota based autopruning in effect, could they conflict with each other?

bcaton85: Yes, that's a great question. They shouldn't conflict with each other since it's essentially the same as two tags being manually deleted at the same time. This is something we'll have to keep in mind during implementation.

enhancements/namespace-auto-pruning.md
bcaton85: We could possibly put something back on the queue to be ran again after a pruning execution but that may inflate the `queueitem` table

bcaton85: @hdonnay I added the `FOR UPDATE SKIP LOCKED` method to the document. This has benefits over the redis lock where we don't have to make multiple requests to find an unlocked row.

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+### Non-Goals
+
+* Allowing users to specify multiple policies per namespace
+* Applying policies based on repo or tag name pattern matching
+* Allowing policy configurations at the repository level
+
+## Proposal
+
+The details in this proposal is only for namespace auto-pruning policies but designed in a way to support repository auto-pruning policies in the future. This proposal is structured in three parts: how policies are managed, how the policies are structured in the DB, and how the policies will be executed via the auto-prune worker.
```
kleesc, line 57: Why did you decide to go for a namespace granularity vs repository when pruning tags? Were there any implementations issue you saw with the latter?

If you can do repository, you could in theory do namespace by going through that all of that namespace's repositories.

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+In order for a container registry to run cost effectively unused resources should be deleted to reclaim storage space. This task should be done automatically in Quay by deleting unused tags through a given policy. The policy used should be configured by the user and a variety policy options should be given. Examples of initial policies would be prune tags older than X number of days or prune tags when more than 100 are present.
+
+## Motivation
+
+Reclaiming space is essential in running a cost efficient registry. Having this done automatically within Quay removes the need for manual intervention or custom software.
+
+### Goals
+
+* User is able to provide a time based policy for auto-pruning tags within a namespace
+* User is able to provide a tag number based policy for auto-pruning tags within a namespace
```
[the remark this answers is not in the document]
bcaton85: This would be a target number of alive tags in a namespace. Say it's configured at 10 tags but 11 exist one tag will be pruned. That tag will be pruned by the oldest creation date.

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+* User is able to specify whether tags go into the time machine or are immediately deleted
```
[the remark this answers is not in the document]
bcaton85: I would say outside for now, but something that can be implemented somewhat soon after. We can re-use the code for expiring tags that was used in the 3.9 quota feature. We may want to clarify that by "immediately deleted" it means "immediately available for garbage collection"

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,170 @@
...
+* The correct tags are deleted when the policy is applied
+* Pruned tags are tracked via usage logs
+
+### Non-Goals
+
+* Allowing users to specify multiple policies per namespace
+* Applying policies based on repo or tag name pattern matching
```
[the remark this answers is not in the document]
bcaton85: Yeah I agree, this design should support it in the future.

enhancements/namespace-auto-pruning.md
berryq460: Great job on outlining the goals of the auto-pruning feature in this pull request! Also, your breakdown into MVP, short-term, and long-term goals provides a clear and structured view of the feature's development trajectory.

"MVP
- Single namespace policies
    - Includes ground work for policy UI, API, and worker

Short term
- Single repository policies
- Multiple namespace policies
- Multiple repository policies

Long term
- Repository name pattern matching for namespace policies
- Tag name pattern matching for repository policies
- Prune untagged artifacts
- System wide auto-pruning configuration
- Dry run policies"

[
] Full roadmap from PM's perspective (https://docs.google.com/document/d/1qA6nxlGmkbVnmA5NtCRYPQERhZlRFEl3LjaODlldbec/edit)

The MVP phase focusing on single namespace policies is a solid starting point. The inclusion of groundwork for policy UI, API, and the worker is a crucial foundation that sets the stage for further development.

Your short-term goals, which include extending to single repository policies and accommodating multiple namespace and repository policies, show a path to the feature's scalability and the potential for varied user scenarios.

The long-term vision is encompassing advanced capabilities like repository name and tag name pattern matching, pruning untagged artifacts, system-wide configuration, and dry run policies. These objectives indicate a commitment to offering a comprehensive and versatile auto-pruning solution over time.

Overall, your pull request presents a well-structured and forward-thinking plan that aligns with the user needs and the feature's growth potential. Your roadmap covers a range of scenarios and considerations, making it a robust foundation for the auto-pruning feature's development. Well done!

enhancements/namespace-auto-pruning.md
flavianmissi: As @BillDett mentioned, proxy organizations already have a form of auto-pruning in place. Should we eventually migrate this proxy specific auto-pruning into one of the options users can enable on a namespace level? The proxy specific auto-pruning is very similar to the age policy.

enhancements/namespace-auto-pruning.md
kleesc: Could this be something that the service-tool would handle?

bcaton85: For Quay.io the service tool could leverage policy enforcement code.

kleesc (approved):

enhancements/namespace-auto-pruning.md
dmesser: I think it makes sense to add this MVP / Short-Term / Long-term perspectives to the proposal. Long-term we also need policies based of "recent pull date:

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Proposal
+
+The details in this proposal is only for namespace auto-pruning policies but designed in a way to support repository auto-pruning policies in the future. This proposal is structured in three parts: how policies are managed, how the policies are structured in the DB, and how the policies will be executed via the auto-prune worker.
+
+### Managing Policies
+
+A user is able to configure auto-pruning within the namespace settings. To start, there will be two auto-pruning options:
+* Prune by number of tags, provided the number of tags
+* Prune tags older than date, provided the number of days
```
dmesser, line 63: Here it would make sense to summarize the policy logic, i.e. "prune by number of tags: gets triggered when the actual number exceeds the desired and then starts deleting according to oldest tag by creation date"

enhancements/namespace-auto-pruning.md
```diff
@@ -0,0 +1,188 @@
...
+* Allowing policy configurations at the repository level
+* User is able to specify whether tags go into the time machine or are immediately deleted
+
+## Proposal
+
+The details in this proposal is only for namespace auto-pruning policies but designed in a way to support repository auto-pruning policies in the future. This proposal is structured in three parts: how policies are managed, how the policies are structured in the DB, and how the policies will be executed via the auto-prune worker.
+
+### Managing Policies
+
+A user is able to configure auto-pruning within the namespace settings. To start, there will be two auto-pruning options:
```
dmesser, line 61: What do you think about defining policies in a more flexible way: 

1) Triggering event (e.g. number of tags exceeds X, or quota is reaching  Y% of its limit or EVERYTIME)
2) Deletion criteria (e.g. tag creation date older than X days, last pull date older than Y days, tag name matches pattern or any combination of them or ALL)

That would create a flexible system to mix and match policies together. Initially we would only start with the following:

1) Supported triggers: number of tags exceeds X, every time
2) Deletion criteria: tags with creation date older than Y days, ALL (sorted by creation or last pull date)

The processing would need to be smart enough to understand that the trigger event defines the limit when to stop deleting, e.g. when number of tags does not exceed X anymore or quota fell below Y%.

The MVP criteria could then be met with: the following selection:

> Prune by number of tags, provided the number days in X:
Trigger event: number of tags exceeds X
Deletion criteria: ALL (sorted by creation date)

> Prune tags older than date, provided number of days in X:
Trigger event: every time (every time the worker wakes up)
Deletion criteria: tag has a creation date older than X days

Initially we could make it so that the UI only offers these two fixed combinations behind a simple dropdown for each policy type. This would leave space for more complex policy creation later.

dmesser (changes requested):

enhancements/namespace-auto-pruning.md
kleesc: We don't currently have a good way of checking a tag's last pull date, short of going through all of the audit logs. Even then, it might not be the case where the entire audit history is being kept. i.e one might be discarding audit logs older that some time range, and you end up losing that information on latest pull time.

bcaton85: I think this creates a more complicated UX and implementation. I think it's better to have unique policy pruning methods where we can enforce what the user can specify rather than account for every combination that a user can provide. Also we're running short on time and accounting for this will put the date at risk.

bcaton85: I think we can also achieve flexibility once we support multiple policies. At that point these same operations can be done.

dmesser: Great feedback. The pull date is something we would do in the future, it was just another criteria as an illustrative example. I agree with @bcaton85 about the complexity of my proposal and the extensibility of the current proposal. Let's go with the current model.

enhancements/namespace-auto-pruning.md
bcaton85: I added more detail

enhancements/namespace-auto-pruning.md
bcaton85: Agreed, added

dmesser (approved):

PR #30, opened 2024-01-30
Proposal for repository enhancement (PROJQUAY-6279)

harishsurf:


dmage:
@bcaton85 can you take a look?

bcaton85:
Yes

enhancements/repositories-auto-pruning.md
```diff
@@ -0,0 +1,179 @@
...
+The existing auto-pruning worker will be refactored to support both namespace and repository level policies. It's workflow is based on values in the `autoprunetaskstatus` table.
+
+**Worker Pseudocode**
+
+* Worker starts on a set interval (1hr)
+* Select a row from `autoprunetaskstatus` with the least or null `last_ran_ms` and `FOR UPDATE SKIP LOCKED`
+    * A null `last_ran_ms` indicates that the task was never ran
+    * A task that hasn't been ran in the longest amount of time or has never been ran at all are prioritized.
+* Get the namespace policy(s) configuration for the namespace from `namespaceautoprunepolicy`
+    * If no namespace policy exists, look up repositories under the namespace containing repository policy(s).
```
bcaton85, line 134: here we only need to check that one repository policy exists. Could probably use peewee's `.exists()`

enhancements/repositories-auto-pruning.md
```diff
@@ -0,0 +1,179 @@
...
+**Worker Pseudocode**
+
+* Worker starts on a set interval (1hr)
+* Select a row from `autoprunetaskstatus` with the least or null `last_ran_ms` and `FOR UPDATE SKIP LOCKED`
+    * A null `last_ran_ms` indicates that the task was never ran
+    * A task that hasn't been ran in the longest amount of time or has never been ran at all are prioritized.
+* Get the namespace policy(s) configuration for the namespace from `namespaceautoprunepolicy`
+    * If no namespace policy exists, look up repositories under the namespace containing repository policy(s). 
+        * If no repository policy exist, delete the entry from `autoprunetaskstatus` for this namespace and exit immediately.
+        * If repository policies exist, skip deleting `autoprunetaskstatus` table entry. Instead loop through all repository policies and apply them sequentially.
```
bcaton85, line 136: Would the policies be executed here or on line 139? It would be better to have it on line 139 to make resolving the policies available in one place.

enhancements/repositories-auto-pruning.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Motivation
+
+Reclaiming space is essential in running a cost efficient registry. Having this done automatically within Quay removes the need for manual intervention or custom software.
+
+### Goals
+
+* User is able to provide a time based policy for auto-pruning tags within a repository.
+* User is able to provide a tag number based policy for auto-pruning tags within a repository.
+* Namespace and repository policy are evaluated and a wider policy is applied.
```
LiZhang19817, line 37: Suppose there's auto-prune policy at Org level as "keeping 10 tags", and there's auto-prune policy at Repo level as "keeping 5 tags", which one will apply to the repo, should be repo level policy, correct?

enhancements/repositories-auto-pruning.md
harishsurf: yes, only in the case when repo policies exists

enhancements/repositories-auto-pruning.md
harishsurf: Yes, always the wider policy which selects more tags to delete is applied. You can refer to the table in the enhancement. I have also tagged you in the google doc which has a table matrix with example values

PR #31, opened 2024-05-06
Trigger notification event for Image expiry (PROJQUAY-7075)

Sunandadadi:


enhancements/trigger-notification-on-image-expiry.md
```diff
@@ -0,0 +1,103 @@
...
+
+Add a new entry to the `ExternalNotificationEvent` table where name = `repo_image_expiry`. This event is used to
+identify `RepositoryNotification` entries that will trigger notifications on image expiry of a repository.
+
+### Image-expiry worker
+
+**Worker Pseudocode**
+
+* Worker starts on a set interval
+* Select a row from `ExternalNotificationEvent` where `event = "repo_image_expiry" & number_of_failures < 3` and `FOR UPDATE SKIP LOCKED`
```
bcaton85, line 70: Do you mean `RepositoryNotification` table instead of `ExternalNotificationEvent` here? 

What will happen if a user tries updating their config while the row is locked?

enhancements/trigger-notification-on-image-expiry.md
```diff
@@ -0,0 +1,103 @@
...
+  * null `eventConfig.last_ran_ms` indicates that the task was never ran
+  * `number_of_failures` indicates failure by the notificationworker queue  
+  * sort query response by `eventConfig.last_ran_ms` asc to prioritize tasks that were never ran or did not run in the longest time
```
bcaton85, line 73: You won't be able to efficiently sort since `last_ran_ms` is within the text field

enhancements/trigger-notification-on-image-expiry.md
```diff
@@ -0,0 +1,103 @@
...
+* Get the count of tags that
+  * belong to the same repository as the configured notificationEvent, and
+  * tag is expiring within configured notificationEvent days
+  * include tags expiry due to org-level/repo-level auto-prune policies
+* If count of tags > 0:
+  * push event into notificationworker queue
+* Else:
+  * Update `event.eventConfig.last_ran_ms` to the current time
```
bcaton85, line 81: How often will the worker push a notification, once when the tag is within the window? How will the worker know it already sent a notification?

enhancements/trigger-notification-on-image-expiry.md
```diff
@@ -0,0 +1,103 @@
...
+* Select a row from `ExternalNotificationEvent` where `event = "repo_image_expiry" & number_of_failures < 3` and `FOR UPDATE SKIP LOCKED`
+  * null `eventConfig.last_ran_ms` indicates that the task was never ran
+  * `number_of_failures` indicates failure by the notificationworker queue  
+  * sort query response by `eventConfig.last_ran_ms` asc to prioritize tasks that were never ran or did not run in the longest time
+* Get the count of tags that
+  * belong to the same repository as the configured notificationEvent, and
+  * tag is expiring within configured notificationEvent days
+  * include tags expiry due to org-level/repo-level auto-prune policies
+* If count of tags > 0:
+  * push event into notificationworker queue
```
bcaton85, line 79: Will a notification be put on the queue for each individual tag? Or a single notification containing a list of tags?

enhancements/trigger-notification-on-image-expiry.md
Sunandadadi: Yes, meant `RepositoryNotification`, thanks for catching it.
Once we create a notification, there is no option to edit the config. It can only be deleted.

enhancements/trigger-notification-on-image-expiry.md
Sunandadadi: good point. I can add a new column for `last_ran_ms` to the `RepositoryNotification` instead.

enhancements/trigger-notification-on-image-expiry.md
Sunandadadi: The worker will push once into the notification queue if tags/tags are expiring in the window. `last_ran_ms` is used to track when the image expiry worker ran for the notification.

enhancements/trigger-notification-on-image-expiry.md
Sunandadadi: One notification with all expiring soon tags.

Sunandadadi: I will add this to the markdown.

enhancements/trigger-notification-on-image-expiry.md
bcaton85: What happens if they try to delete it when the row is locked?

enhancements/trigger-notification-on-image-expiry.md
bcaton85: Yes but the worker will have to run multiple times to check if there are any new tags that are about to expire right? How will it know that it sent a notification for a specific tag in the last run vs the next run?

enhancements/trigger-notification-on-image-expiry.md
Sunandadadi: > What happens if they try to delete it when the row is locked?

I would guess that the delete should fail with error unable to acquire lock. But, will test this out to be sure.

enhancements/trigger-notification-on-image-expiry.md
```diff
@@ -0,0 +1,128 @@
...
+  * null `last_ran_ms` indicates that the task was never ran
+  * `number_of_failures` indicates failure by the notificationworker queue  
+  * sort query response by `last_ran_ms` asc to prioritize tasks that were never ran or did not run in the longest time
+* Get the count of tags that
+  * belong to the same repository as the configured notificationEvent, and
+  * tag is expiring within configured notificationEvent days
+  * include tags expiry due to org-level/repo-level auto-prune policies
+  * tag does not belong to `tagnotificationsent` table for the notification method
+* If count of tags > 0:
+  * push event into notificationworker queue
```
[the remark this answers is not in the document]
Sunandadadi: yes, I will update the md.

enhancements/trigger-notification-on-image-expiry.md
```diff
@@ -0,0 +1,128 @@
...
+
+Add a new entry to the `ExternalNotificationEvent` table where name = `repo_image_expiry`. This event is used to
+identify `RepositoryNotification` entries that will trigger notifications on image expiry of a repository.
+
+A new table will be added to track tags that were successfully notified.
+
+**tagNotificationSent**
+
+A new entry is added to this table when a notification was sent for a tag. In the future runs of the imageexpiry worker,
+this table is used to filter remaining tags for a repository that need to be notified.
```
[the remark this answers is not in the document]
Sunandadadi: good question, I think this table should be cleaned when a tag expires or when the expiry of the tag is updated.

enhancements/trigger-notification-on-image-expiry.md
```diff
@@ -0,0 +1,135 @@
...
+
+| Field | Datatype | Attributes | Description|
+| --- | ----------- | ----------- | ----------- |
+| id | integer | Unique, not null, auto increment | Unique identifier for the table |
+| notification_id | integer | FK, not null | RepositoryNotification.id |
+| tag_id | integer | FK, not null | Tag.id |
```
bcaton85, line 76: `notification_id` and `tag_id` will have to be indexed

bcaton85 (approved):
LGTM

PR #32, opened 2024-06-10
enhancement: manifest annotations

Marcusk19:
Enhancement proposal for separating manifest annotations from labels in quay

### DIFF
--- a/enhancements/annotation-storage.md
+++ b/enhancements/annotation-storage.md
@@ -1,0 +1,126 @@
+---
+title: Annotation-Querying
+authors:
+  - "@mkok"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2024-06-07
+last-updated: yyyy-mm-dd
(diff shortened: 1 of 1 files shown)


enhancements/annotation-storage.md
```diff
@@ -0,0 +1,126 @@
...
+
+A new endpoint will need to be added that allows users to access manifest annotations.
+It should be similar to the labels endpoint to avoid confusion.
+
+**Get annotation(s)**
+```
+GET /api/v1/repository/{repository}/manifest/{manifestref}/annotations
+```
+
+**Update annotation**
```
syed, line 92: annotations can be in multiple places in the manifest. The most obvious one is  at the top layer `.annotations` but there also be annotations on each layer and also in the `config` we should provide the ability to query/update them as well

Marcusk19: Ok will those also be part of the `manifest_bytes` column?

PR #40, opened 2026-04-17
PROJQUAY-5850: feat(operator): standardized STS configuration via OLM and CCO

quay-devel:
## Summary
- Design proposal for integrating the Quay operator with the standardized OpenShift STS configuration flow using OLM and CCO (`CredentialRequest` API) per OCPSTRAT-171 / OCPSTRAT-6
- Enables Quay on STS-enabled OpenShift clusters (ROSA, OSD) to authenticate to AWS S3 using short-lived STS credentials instead of static IAM keys
- Applies only to `ObjectStorage: managed: false` (unmanaged, real AWS S3); managed NooBaa storage is unaffected

## Root Cause / Rationale
- ROSA/OSD clusters prohibit static IAM user keys; customers using real AWS S3 with `managed: false` have no supported path today
- OCPSTRAT-6 mandates all OLM operators capable of cloud API integration adopt the CCO-based `CredentialRequest` flow
- Strategic customer (Elevance Health/Anthem) has this as a hard requirement for Quay on ROSA migration

## Changes
- `enhancements/sts-olm-cco-support.md` — new enhancement proposal covering:
  - Opt-in via `ROLEARN` env var in OLM Subscription; graceful fallback when absent
  - `CredentialRequest` CR lifecycle with `cloudTokenPath` and `serviceAccountNames` fields
  - CSV changes: `token-auth-aws` annotation, `bound-sa-token` projected volume, cluster-scoped RBAC
  - STS-capable cluster detection (platform type, CCO mode, CRD availability)
  - Credential conflict detection when static keys coexist with `ROLEARN`
  - `RolloutBlocked` condition when CCO fails to reconcile
  - Required IAM permissions derived from boto3 call-site analysis
  - RHEL-based deployment guidance (instance profile, `STSS3Storage`, web identity token)
  - Open design questions: runtime vs bundle-shipped vs direct Secret creation approaches
  - Test plan (unit, integration, E2E kuttl on ROSA), upgrade/downgrade strategy

## Test Plan
- [ ] Unit tests added/updated
- [ ] Integration/registry tests pass
- [ ] Type checking passes
- [ ] Manual testing performed
- [ ] Frontend tests pass
- [ ] Migration tested (upgrade and downgrade)

N/A — this is an enhancement proposal document, not a code change. Test plan for the implementation is defined within the enhancement.

## JIRA
https://redhat.atlassian.net/browse/PROJQUAY-5850

## Backport
- [ ] Backport required
- [x] No backport needed

### DIFF
--- a/enhancements/sts-olm-cco-support.md
+++ b/enhancements/sts-olm-cco-support.md
@@ -1,13 +1,16 @@
 ---
 title: Standardized STS Configuration via OLM and CCO for Quay on OpenShift
 authors:
-  - TBD
+  - "@dmesser"
+  - "@doconnor"
+  - "@tlwu2013"
 reviewers:
-  - TBD
+  - "@jbpratt"
+  - "@tlwu2013"
 approvers:
   - TBD
 creation-date: 2023-07-19
-last-updated: 2023-07-19
+last-updated: 2026-05-21
 status: implementable
 see-also:
   - "https://issues.redhat.com/browse/OCPSTRAT-171"
@@ -42,6 +45,20 @@ When `ObjectStorage` is set to `managed: false`, the customer provides their own
 
 Red Hat's platform strategy (OCPSTRAT-6) mandates that all OLM-managed operators capable of integrating with cloud-provider APIs adopt the CCO-based `CredentialRequest` flow. Quay has been identified as a target operator. A strategic customer (Elevance Health/Anthem) has this as a hard requirement for migrating Quay to ROSA.
 
+### Why the Operator Manages Credentials for Unmanaged Storage
+
+When `ObjectStorage: managed: false`, the operator does not create or manage the S3 bucket — the customer does. So why should the operator manage cloud credentials?
+
+The answer is that the operator is not managing *IAM credentials* in the traditional sense (it never creates IAM users, keys, or policies). Instead, it acts as a **configuration broker** between the OCP platform and the Quay application pods. Specifically:
+
+1. **The customer creates an IAM role** with the necessary S3 permissions and trust policy. This is the customer's responsibility, just like creating the bucket.
+2. **The customer supplies the role ARN** via the OLM Subscription (`ROLEARN`), following the same standardized pattern used by every other OLM operator that supports STS (OADP, cert-manager, etc.).
+3. **The operator creates a `CredentialRequest`** — a declarative request that tells CCO "the `quay-app` service account needs to assume this role." The operator does not generate, store, or rotate any credentials itself.
+4. **CCO provisions a credentials file** (containing the role ARN and the OIDC token path) and stores it in a Secret. This file is a *configuration pointer*, not a credential — it tells boto "call `AssumeRoleWithWebIdentity` with this role using the projected OIDC token."
+5. **The operator mounts this Secret** into the Quay application pods and sets `AWS_SHARED_CREDENTIALS_FILE`. This is the same kind of configuration injection the operator already performs (TLS certificates, config bundles, etc.).
+
+The operator already manages the `quay-app` Deployment, ServiceAccount, and volumes. Adding one more volume mount for CCO-provisioned config is consistent with its existing role. The alternative — asking customers to manually create Secrets, add volume mounts, and set environment variables — would break the standardized OCPSTRAT-171 UX that other operators provide and that OperatorHub surfaces.
+
 ### Goals
 
 - Implement the standardized CCO `CredentialRequest` flow for the `quay-app` service account when `ObjectStorage: managed: false` and the cluster is STS-capable.
@@ -72,9 +89,11 @@ Understanding the credential flow is essential because this is NOT the tradition
    OLM injects ROLEARN into all pods managed by this operator (including quay-app pods)
 
 2. Operator reads ROLEARN, detects STS-capable cluster, creates CredentialRequest
-   with serviceAccountNames: [quay-app] and stsIAMRoleARN: <ROLEARN value>
+   with serviceAccountNames: [quay-app], stsIAMRoleARN: <ROLEARN value>,
+   and cloudTokenPath: /var/run/secrets/openshift/serviceaccount/token
    ↓
-   CCO provisions a Secret containing a credentials file:
+   CCO validates the operator's OIDC token at cloudTokenPath,
+   then provisions a Secret containing a credentials file:
 
      [default]
      sts_regional_endpoints = regional
@@ -207,9 +226,10 @@ spec:
     namespace: <quayregistry-namespace>
   serviceAccountNames:
     - quay-app
+  cloudTokenPath: /var/run/secrets/openshift/serviceaccount/token
 ```
 
-`stsIAMRoleARN` (available since OCP 4.14 CCO) tells CCO to produce a web-identity credentials file rather than static IAM user keys. `serviceAccountNames: [quay-app]` is a required enforcement field — CCO rejects CredentialRequests without it. The `statementEntries` use `resource: "*"` because the operator does not know the customer's bucket name when storage is unmanaged; the actual bucket-scoped IAM policy is the customer's responsibility when creating the role.
+`stsIAMRoleARN` (available since OCP 4.14 CCO) tells CCO to produce a web-identity credentials file rather than static IAM user keys. `serviceAccountNames: [quay-app]` is a required enforcement field — CCO rejects CredentialRequests without it. `cloudTokenPath` tells CCO where the operator pod's projected OIDC token is located; CCO uses this to verify the operator's identity when processing the `CredentialRequest`. The `statementEntries` use `resource: "*"` because the operator does not know the customer's bucket name when storage is unmanaged; the actual bucket-scoped IAM policy is the customer's responsibility when creating the role.
 
 CCO produces the Secret `<quayregistry-name>-quay-app-aws` containing:
 
@@ -344,8 +364,50 @@ Derived from static analysis of all boto3 call sites in `storage/cloud.py` (`qua
 
 ### CSV Changes
 
+Three additions are required in the ClusterServiceVersion:
+
+#### Annotation
+
 ```yaml
 features.operators.openshift.io/token-auth-aws: "true"   # changed from "false"
+```
+
+#### Projected OIDC Token Volume for the Operator Pod
+
+The operator pod needs a projected `bound-sa-token` volume so that its own OIDC token is available at the path specified by `cloudTokenPath` in the `CredentialRequest`. Without this volume, the directory `/var/run/secrets/openshift/serviceaccount/` does not exist in the operator pod and CCO cannot validate the `CredentialRequest`.
+
+Add to the operator Deployment spec in the CSV:
+
+```yaml
+spec:
+  template:
+    spec:
+      containers:
+        - name: quay-operator
+          volumeMounts:
+            - name: bound-sa-token
+              mountPath: /var/run/secrets/openshift/serviceaccount
+              readOnly: true
+      volumes:
+        - name: bound-sa-token
+          projected:
+            sources:
+              - serviceAccountToken:
+                  path: token
+                  audience: openshift
+```
+
+#### Additional RBAC
+
+The operator needs read access to cluster infrastructure and cloud credential configuration for STS-capability detection (already listed in section 2 above, repeated here for CSV completeness):
+
+```yaml
+- apiGroups: ["config.openshift.io"]
+  resources: ["infrastructures"]
+  verbs: ["get"]
+- apiGroups: ["operator.openshift.io"]
+  resources: ["cloudcredentials"]
+  verbs: ["get"]
 ```
 
 ### RHEL-Based Quay Deployments
@@ -380,6 +442,30 @@ DISTRIBUTED_STORAGE_CONFIG:
 | CredentialRequest rejected by CCO (missing `serviceAccountNames`) | CCO 4.14+ requires this field; operator always populates it |
 | Regression on non-STS upgrades | STS path requires `ROLEARN` env var; existing Subscriptions without it are fully unaffected |
 | Multipart upload in-flight when OIDC token rotates | boto re-fetches the token file on each credential refresh cycle; the token at the path is updated by kubelet before expiry |
+
+### Open Design Questions
+
+#### Bundle-Shipped vs Runtime-Created CredentialRequest
+
+This enhancement proposes creating the `CredentialRequest` at runtime during reconciliation. An alternative approach (outlined in PROJQUAY-5850 comments) is to ship the `CredentialRequest` manifest in the operator bundle under `manifests/` so that `ccoctl` or CCO can process it during operator install.
+
+**Bundle-shipped** (standard OCPSTRAT-171 pattern, used by OADP):
+- Pros: Aligns with the documented CCO flow; `ccoctl` can extract and pre-provision credentials during disconnected installs; simpler operator code.
+- Cons: The CredentialRequest is static — cannot vary `stsIAMRoleARN` per QuayRegistry; requires the role ARN at bundle-build time or a fixed environment variable substitution.
+
+**Runtime-created** (used by some newer operators):
+- Pros: Can construct the CredentialRequest dynamically per QuayRegistry; handles the role ARN from `ROLEARN` env var naturally; supports multiple QuayRegistry instances with different roles in the future.
+- Cons: More operator code; requires RBAC for CredentialRequest CRUD; does not integrate with `ccoctl` disconnected install flow.
+
+**Recommendation**: Start with the bundle-shipped approach for OCPSTRAT-171 conformance. The `ROLEARN` value can be injected via OLM Subscription env vars, which the standard flow already supports. If per-registry role ARN support is needed later, runtime creation can be added as a follow-on.
+
+#### Storage Driver: `S3Storage` with Web Identity vs `STSS3Storage`
+
+The `STSS3Storage` driver in `quay/quay` uses `sts:AssumeRole` with static IAM user keys to obtain temporary credentials. This is a different STS flow from the CCO/web-identity path.
+
+With the CCO approach, the credentials file provisioned by CCO contains `role_arn` and `web_identity_token_file`. When `AWS_SHARED_CREDENTIALS_FILE` is set, boto3's standard credential chain resolves this automatically via `AssumeRoleWithWebIdentity` — no Quay code changes needed. The standard `S3Storage` driver works as-is because boto handles credential resolution transparently.
+
+**Recommendation**: Use `S3Storage` (not `STSS3Storage`) for the CCO path. `STSS3Storage` remains available for RHEL-based deployments where cross-account assume-role with static keys is the only option.
 
 ## Design Details
 
@@ -429,6 +515,10 @@ The `CredentialRequest` CRD is provided by CCO, which ships with OCP. The operat
 ## Implementation History
 
 - 2023-07-19 PROJQUAY-5850 filed; feasibility investigation completed.
+- 2026-04-16 Targeted for Quay 3.19 (Q4 2026); GCS WIF (PROJQUAY-7729) to share operator-side pattern.
+- 2026-04-17 Initial enhancement PR opened; review feedback on authors, credential brokering rationale.
+- 2026-04-28 Review feedback: `cloudTokenPath` and `bound-sa-token` volume additions required.
+- 2026-05-21 Enhancement updated to address review feedback; open design questions documented.
 
 ## Drawbacks
 
--- a/enhancements/sts-olm-cco-support.md
+++ b/enhancements/sts-olm-cco-support.md
@@ -188,7 +188,9 @@ For reference, the OADP operator's CSV already uses `clusterPermissions` with `c
 
 #### 3. CredentialRequest — Created at Runtime
 
-The operator creates the `CredentialRequest` programmatically during reconciliation (not packaged in the bundle — OKD documentation explicitly states that bundled CredentialRequests are not supported). One `CredentialRequest` is created per `QuayRegistry` in the registry's namespace, owned by the `QuayRegistry` for garbage collection:
+The operator creates the `CredentialRequest` programmatically during reconciliation (not packaged in the bundle — OKD documentation explicitly states that bundled CredentialRequests are not supported). One `CredentialRequest` is created per `QuayRegistry` in the registry's namespace, owned by the `QuayRegistry` for garbage collection.
+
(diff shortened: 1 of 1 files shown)


enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,277 @@
+---
+title: Standardized STS Configuration via OLM and CCO for Quay on OpenShift
+authors:
+  - "@dmesser"
+  - "@doconnor"
```
jbpratt, line 5: Authors are not accurate

quay-devel: Fixed — removed the Jira reporter/assignee handles which were not necessarily the document authors. Set to TBD pending the correct GitHub handles from the team.

quay-devel: Fixed — removed the Jira reporter/assignee handles which were not necessarily the document authors. Set to TBD pending the correct GitHub handles from the team.

jbpratt (changes requested):
Hmm this doesn't make sense yet.. the operator doesn't manage the aws s3 bucket (unmanaged object store) so why should it manage the credentials and what should that look like?

enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,448 @@
...
+          - s3:HeadBucket
+          - s3:GetBucketLocation
+          - s3:GetBucketCors
+          - s3:PutBucketCors
+        resource: "*"
+  secretRef:
+    name: <quayregistry-name>-quay-app-aws
+    namespace: <quayregistry-namespace>
+  serviceAccountNames:
+    - quay-app
```
tlwu2013, line 235: The `cloudTokenPath` is missing from the spec (w/o it, CCO cannot verify the operator's identity).

The guide ("How To Short-Lived Token Auth", step 3 for AWS) requires setting `cloudTokenPath: /var/run/secrets/openshift/serviceaccount/token`, so CCO knows where to find the operator pod's projected OIDC token when processing the `CredentialRequest`. 
https://docs.google.com/document/d/1iFNpyycby_rOY1wUew-yl3uPWlE00krTgr9XHDZOTNo/edit?tab=t.0#heading=h.ugf6s9d5c2k2

```
  serviceAccountNames:
    - quay-app
  cloudTokenPath: /var/run/secrets/openshift/serviceaccount/token   # add this
```

enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,448 @@
...
+      }
+    }
+  ]
+}
+```
+
+### CSV Changes
+
+```yaml
+features.operators.openshift.io/token-auth-aws: "true"   # changed from "false"
```
tlwu2013, line 393: If I read the guide correctly, it also mentions two additional CSV changes for the operator pod alongside the annotation:
- A `bound-sa-token` projected volume mount so the operator itself has an OIDC token at `/var/run/secrets/openshift/serviceaccount/token` <--this is what `cloudTokenPath` in the `CredentialRequest` points to.

- RBAC for `config.openshift.io/infrastructures` and `operator.openshift.io/cloudcredentials`.

This volume is needed for the operator pod's `cloudTokenPath` directory to exist.  CCO cannot validate the `CredentialRequest` without it. 

```
# Add to operator Deployment in CSV:
volumeMounts:
  - name: bound-sa-token
    mountPath: /var/run/secrets/openshift/serviceaccount
    readOnly: true
volumes:
  - name: bound-sa-token
    projected:
      sources:
        - serviceAccountToken:
            path: token
            audience: openshift
```

quay-devel:
## Addressed Review Feedback (a560fb7)

### @jbpratt — "why should it manage the credentials?"

Added a new section **"Why the Operator Manages Credentials for Unmanaged Storage"** that directly addresses this concern. TL;DR: the operator is not managing *IAM credentials* — it acts as a **configuration broker**. It creates a `CredentialRequest` (a declarative request, not a credential), CCO provisions a credentials file (a config pointer telling boto where to find the OIDC token), and the operator mounts that file into quay-app pods. This is the same pattern as OADP and cert-manager, and consistent with how the operator already injects TLS certs and config bundles.

Also added **Open Design Questions** surfacing two decisions for the team:
1. **Bundle-shipped vs runtime-created CredentialRequest** — recommending bundle-shipped for OCPSTRAT-171 conformance
2. **S3Storage vs STSS3Storage** — recommending `S3Storage` for the CCO path (boto resolves web-identity credentials transparently)

### @jbpratt — "Authors are not accurate"

Updated to `@dmesser`, `@doconnor`, `@tlwu2013` based on JIRA context. Please correct if needed.

### @tlwu2013 — missing `cloudTokenPath`

Added `cloudTokenPath: /var/run/secrets/openshift/serviceaccount/token` to the `CredentialRequest` spec and updated the credential flow diagram to show CCO validating the operator's OIDC token at this path.

### @tlwu2013 — missing CSV changes (bound-sa-token volume + RBAC)

Expanded the **CSV Changes** section with three subsections:
1. **Annotation** (`token-auth-aws: "true"`)
2. **Projected OIDC Token Volume** — `bound-sa-token` projected volume mount for the operator pod
3. **Additional RBAC** — `config.openshift.io/infrastructures` and `operator.openshift.io/cloudcredentials` get access

enhancements/sts-olm-cco-support.md
quay-devel: Updated in a560fb7 — set authors to @dmesser, @doconnor, @tlwu2013 based on JIRA ticket reporter/assignee and active reviewers. Please correct if these aren't the right handles.

enhancements/sts-olm-cco-support.md
quay-devel: Good catch — added in a560fb7. The `CredentialRequest` spec now includes:

```yaml
serviceAccountNames:
  - quay-app
cloudTokenPath: /var/run/secrets/openshift/serviceaccount/token
```

Also updated the credential flow diagram (step 2) to show CCO validating the operator's OIDC token at `cloudTokenPath` before provisioning the Secret.

enhancements/sts-olm-cco-support.md
quay-devel: Both additions incorporated in a560fb7. The **CSV Changes** section now covers three items:

1. **Annotation**: `token-auth-aws: "true"`
2. **Projected OIDC Token Volume**: `bound-sa-token` projected volume with `serviceAccountToken` source (`audience: openshift`) mounted at `/var/run/secrets/openshift/serviceaccount` on the operator pod
3. **Additional RBAC**: `get` on `config.openshift.io/infrastructures` and `operator.openshift.io/cloudcredentials`

This ensures the operator pod has the OIDC token that `cloudTokenPath` points to, and the permissions to detect STS capability.

quay-devel:
@jbpratt re: "the operator doesn't manage the aws s3 bucket (unmanaged object store) so why should it manage the credentials and what should that look like?"

Addressed in a560fb7 with a new section **"Why the Operator Manages Credentials for Unmanaged Storage"**. The short version:

The operator is **not** managing IAM credentials — it never creates IAM users, keys, or policies. It acts as a **configuration broker**: it creates a `CredentialRequest` (a declarative ask to CCO), CCO provisions a credentials *file* (which is really a config pointer — `role_arn` + `web_identity_token_file`), and the operator mounts that file into quay-app pods. This is the same kind of volume injection it already does for TLS certs and config bundles.

The customer still owns the IAM role, the trust policy, and the bucket — exactly as they do today. The operator just bridges the gap between the OLM Subscription (where the admin supplies `ROLEARN`) and the Quay app pods (which need `AWS_SHARED_CREDENTIALS_FILE` set).

This is the standard OCPSTRAT-171 pattern used by OADP, cert-manager, and other OLM operators. Without it, admins would need to manually create Secrets, add volume mounts, and set env vars — which breaks the OperatorHub STS installation UX.

Also added an **Open Design Questions** section surfacing two decisions for the team:
1. Bundle-shipped vs runtime-created CredentialRequest (recommending bundle-shipped for conformance)
2. S3Storage vs STSS3Storage for the CCO path (recommending S3Storage — boto resolves web-identity transparently)

enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,538 @@
...
+
+**Bundle-shipped** (standard OCPSTRAT-171 pattern, used by OADP):
+- Pros: Aligns with the documented CCO flow; `ccoctl` can extract and pre-provision credentials during disconnected installs; simpler operator code.
+- Cons: The CredentialRequest is static — cannot vary `stsIAMRoleARN` per QuayRegistry; requires the role ARN at bundle-build time or a fixed environment variable substitution.
+
+**Runtime-created** (used by some newer operators):
+- Pros: Can construct the CredentialRequest dynamically per QuayRegistry; handles the role ARN from `ROLEARN` env var naturally; supports multiple QuayRegistry instances with different roles in the future.
+- Cons: More operator code; requires RBAC for CredentialRequest CRUD; does not integrate with `ccoctl` disconnected install flow.
+
+**Recommendation**: Start with the bundle-shipped approach for OCPSTRAT-171 conformance. The `ROLEARN` value can be injected via OLM Subscription env vars, which the standard flow already supports. If per-registry role ARN support is needed later, runtime creation can be added as a follow-on.
```
bcaton85: **This recommendation contradicts the implementation details.**

Line 186 states: "The operator creates the CredentialRequest programmatically during reconciliation (not packaged in the bundle)." This recommendation says the opposite: "Start with the bundle-shipped approach."

These are mutually exclusive designs. The entire "CredentialRequest — Created at Runtime" section (Section 3, lines 184-230) describes runtime creation in detail — if bundle-shipped is chosen, that section needs a complete rewrite.

Additionally, the claim that OADP uses bundle-shipped CredentialRequests (line 452) is incorrect. OADP takes a third approach entirely: it reads ROLEARN from the environment, then creates the AWS credential Secret directly — bypassing CCO's CredentialRequest flow. See OADP's `pkg/credentials/stsflow/stsflow.go`. OADP has RBAC for credentialsrequests in its CSV but doesn't use it.

The team actually has three viable approaches:

| Approach | OADP precedent? | ccoctl support | Per-registry ARN | Complexity |
|---|---|---|---|---|
| Runtime CredentialRequest (what Section 3 describes) | No | No | Yes | Medium |
| Bundle-shipped CredentialRequest (what this line recommends) | No | Yes | No | Low |
| Direct Secret creation (what OADP actually does) | Yes | No | Yes | Lowest |

Please: (a) pick one approach, (b) rewrite the contradicting section, (c) acknowledge the OADP shortcut as a viable option.

enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,538 @@
...
+Note: the storage type becomes `S3Storage` (not `RHOCSStorage`). `RHOCSStorage` is only used for managed NooBaa storage; for real AWS S3 the customer's unmanaged config already specifies the correct storage type.
+
+The operator does not generate the storage configuration for unmanaged storage — that comes from the customer's `configBundleSecret`. The operator only ensures that `AWS_SHARED_CREDENTIALS_FILE` is set on the pods. No modification of the customer's `config.yaml` content is needed or performed.
+
+#### 6. Degraded Condition
+
+If the `CredentialRequest` has not reached `status.provisioned == true` within a configurable timeout (default: 5 minutes) after `ROLEARN` is detected, the operator sets:
+
+```
+type:    Degraded
```
bcaton85: **The `Degraded` condition type does not exist in the quay-operator.**

Verified in the quay-operator codebase: the operator uses `Available`, `RolloutBlocked`, and `ComponentsCreated` as condition types. There is no `Degraded` condition defined.

This is also referenced in User Story 3 (line 135) and Dev Preview graduation criteria (line 480).

Options:
1. Define a new `Degraded` condition type and document how it interacts with the existing status controller and `Available` condition aggregation.
2. Reuse the existing `RolloutBlocked` condition with a new reason like `CredentialRequestNotProvisioned`.

Either way, the choice needs to be explicit here since it affects the status controller implementation.

enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,538 @@
...
+
+Before creating a `CredentialRequest`, the operator confirms the cluster is STS-capable. On each reconcile, in order:
+
+1. **`ROLEARN` present**: If `os.Getenv("ROLEARN")` is empty, skip STS — no further checks needed.
+2. **ObjectStorage unmanaged**: If `ComponentObjectStorage` is `managed: true`, skip STS (NooBaa path needs no STS). Log a warning if `ROLEARN` is set with managed storage to alert the admin.
+3. **Platform type**: Read `config.openshift.io/v1 Infrastructure cluster`; confirm `status.platformStatus.type == "AWS"`.
+4. **CCO mode**: Read `operator.openshift.io/v1 CloudCredential cluster`; confirm `spec.credentialsMode` is not `Mint` or `Passthrough` (those modes produce static keys, not web-identity config). Empty `credentialsMode` on AWS means STS mode.
+5. **CRD availability**: Confirm `credentialsrequests.cloudcredential.openshift.io` CRD exists via API discovery. Absent on OCP < 4.14 or non-OCP environments.
+
+New RBAC rules required in CSV:
```
bcaton85: **Blocker: These RBAC rules must be specified as `clusterPermissions`, not namespace-scoped `permissions`.**

`config.openshift.io/infrastructures` and `operator.openshift.io/cloudcredentials` are cluster-scoped resources (there is a single `Infrastructure` object named `cluster` and a single `CloudCredential` object named `cluster`). They cannot be accessed via namespace-scoped RBAC.

Currently (verified), the quay-operator CSV uses ONLY namespace-scoped `permissions` — there are no `clusterPermissions` defined. Adding `clusterPermissions` means OLM will create a ClusterRole and ClusterRoleBinding for the operator SA. This is a material change to the operator's OLM security footprint and should be explicitly discussed.

For reference, the OADP operator's CSV already uses `clusterPermissions` with `cloudcredential.openshift.io/credentialsrequests` RBAC, confirming this is the correct approach.

Please update this section to explicitly state these are `clusterPermissions` and document the security trade-off of a namespace-scoped operator gaining cluster-scoped access.

enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,538 @@
...
+spec:
+  channel: stable-3.13
+  name: quay-operator
+  config:
+    env:
+      - name: ROLEARN
+        value: "arn:aws:iam::123456789012:role/quay-s3-role"
+```
+
+OLM propagates `ROLEARN` as an environment variable into all Deployments managed by the operator — both the operator pod and the Quay application pods. The operator reads `os.Getenv("ROLEARN")` during reconciliation. If it is empty, the STS path is skipped entirely.
```
bcaton85: **Verify: This claim about OLM env var propagation may be inaccurate.**

OLM's `spec.config.env` on a Subscription injects environment variables into the **operator Deployment only** — not into workloads that the operator itself creates. The quay-app pods are created by the operator's reconcile loop, not by OLM.

If OLM only injects into the operator pod, then `ROLEARN` is available via `os.Getenv("ROLEARN")` in the operator's reconcile loop (which is sufficient — the operator reads it to construct the CredentialRequest), but it is NOT automatically present in quay-app pods.

The credential flow diagram at line 89 also states "OLM injects ROLEARN into all pods managed by this operator (including quay-app pods)" — if this is inaccurate, both locations need correction. The `ROLEARN` value only needs to be in the operator pod; quay-app pods get their credentials via the mounted CCO Secret, not via ROLEARN.

enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,538 @@
...
+      s3_bucket: <customer-provided>
+      s3_region: <customer-provided>
+      storage_path: /datastorage/registry
+      # No aws_access_key_id or aws_secret_access_key
+      # boto resolves via AWS_SHARED_CREDENTIALS_FILE → AssumeRoleWithWebIdentity
+```
+
+Note: the storage type becomes `S3Storage` (not `RHOCSStorage`). `RHOCSStorage` is only used for managed NooBaa storage; for real AWS S3 the customer's unmanaged config already specifies the correct storage type.
+
+The operator does not generate the storage configuration for unmanaged storage — that comes from the customer's `configBundleSecret`. The operator only ensures that `AWS_SHARED_CREDENTIALS_FILE` is set on the pods. No modification of the customer's `config.yaml` content is needed or performed.
```
bcaton85, line 282: **Missing: Credential conflict detection needed.**

If a customer's existing `configBundleSecret` contains `s3_access_key` / `s3_secret_key` in `DISTRIBUTED_STORAGE_CONFIG`, boto3 will use those static credentials **instead of** the CCO-provisioned credentials file (explicit credentials take priority in boto's chain). The STS pathway silently does nothing.

The proposal should:
1. Explicitly document that when `ROLEARN` is set, any static credentials in `DISTRIBUTED_STORAGE_CONFIG` must be removed.
2. Have the operator warn or error if both `ROLEARN` is set AND `s3_access_key`/`s3_secret_key` are present in the config bundle.

This is also a missing test case — the Test Plan section has no coverage for this conflict scenario.

enhancements/sts-olm-cco-support.md
```diff
@@ -0,0 +1,538 @@
...
+| Regression on non-STS upgrades | STS path requires `ROLEARN` env var; existing Subscriptions without it are fully unaffected |
+| Multipart upload in-flight when OIDC token rotates | boto re-fetches the token file on each credential refresh cycle; the token at the path is updated by kubelet before expiry |
+
+### Open Design Questions
+
+#### Bundle-Shipped vs Runtime-Created CredentialRequest
+
+This enhancement proposes creating the `CredentialRequest` at runtime during reconciliation. An alternative approach (outlined in PROJQUAY-5850 comments) is to ship the `CredentialRequest` manifest in the operator bundle under `manifests/` so that `ccoctl` or CCO can process it during operator install.
+
+**Bundle-shipped** (standard OCPSTRAT-171 pattern, used by OADP):
```
bcaton85: **Inaccurate: OADP does not use bundle-shipped CredentialRequests.**

OADP reads `ROLEARN` from the environment, then creates the AWS credential Secret directly via `pkg/credentials/stsflow/stsflow.go` — bypassing CCO's CredentialRequest flow entirely. OADP has RBAC for credentialsrequests in its CSV but doesn't actually create them at runtime or ship them in the bundle.

Similarly, cert-manager does not create CredentialRequests — it relies on admins to create them manually and consumes the resulting Secret.

This line should be corrected to avoid citing OADP as precedent for an approach it doesn't use.

enhancements/sts-olm-cco-support.md
quay-devel: You're right on all counts — addressed in fee2f3d.

**Contradiction fixed**: Removed the recommendation to use bundle-shipped approach. The implementation details (Section 3) remains runtime creation as the proposed approach. The Open Design Questions section now presents all three approaches in a neutral comparison table without contradicting the proposal.

**OADP precedent corrected**: Acknowledged that OADP creates the AWS credential Secret directly via `pkg/credentials/stsflow/stsflow.go`, bypassing CCO's CredentialRequest flow. Cert-manager citation also removed. The table now accurately attributes the direct Secret creation approach to OADP.

The recommendation now notes that runtime CredentialRequest (Approach A) is proposed because it delegates to CCO, but Approach C (the OADP pattern) is lowest-complexity with production precedent. Team decides.

enhancements/sts-olm-cco-support.md
quay-devel: Fixed in fee2f3d — replaced all `Degraded` references with `RolloutBlocked`.

The section now explicitly states the quay-operator's actual condition types (`Available`, `RolloutBlocked`, `ComponentsCreated`) and documents how the existing status controller aggregation applies: `Available` remains `False` while any `RolloutBlocked` condition is `True`.

The condition uses reason `CredentialRequestNotProvisioned` under `RolloutBlocked`. This also applies to the new credential conflict case (reason `ConflictingCredentials`).

enhancements/sts-olm-cco-support.md
quay-devel: Fixed in fee2f3d. Both RBAC sections (Section 2 and CSV Changes) now explicitly specify these as `clusterPermissions`, not namespace-scoped `permissions`.

Added documentation of the security trade-off: the quay-operator CSV currently uses only namespace-scoped `permissions`, so adding `clusterPermissions` means OLM will create a ClusterRole and ClusterRoleBinding. This is a material change to the operator's OLM security footprint and should be reviewed by the security team. OADP's CSV is cited as precedent for this being the correct approach.

enhancements/sts-olm-cco-support.md
quay-devel: Good catch — fixed in fee2f3d. You're correct that OLM's `spec.config.env` only injects into the operator Deployment, not into workloads the operator creates.

Both locations updated:
- The credential flow diagram (step 1) now reads: "OLM injects ROLEARN into the operator Deployment only (not into quay-app pods)"
- Section 1 now explains that `ROLEARN` is available via `os.Getenv("ROLEARN")` in the operator's reconcile loop (which is sufficient — the operator reads it to construct the CredentialRequest). The quay-app pods get their credentials via the mounted CCO Secret and `AWS_SHARED_CREDENTIALS_FILE`, not via `ROLEARN`.
- The Goals section no longer claims OLM propagates to "all operator-managed pods."

enhancements/sts-olm-cco-support.md
quay-devel: Agreed — this was a gap. Fixed in fee2f3d.

Added a **Credential conflict detection** subsection under Section 5 (config.yaml Changes):
- When `ROLEARN` is set, the operator must inspect `configBundleSecret` for static `s3_access_key`/`s3_secret_key` in the storage config
- If both present: `RolloutBlocked` condition with reason `ConflictingCredentials` and a message instructing the customer to remove the static keys
- Operator does not roll out quay-app pods until the conflict is resolved

Also added:
- A row in the Risks and Mitigations table
- A new unit test case in the Test Plan: `ROLEARN` set + static keys present → `RolloutBlocked` with `ConflictingCredentials`

enhancements/sts-olm-cco-support.md
quay-devel: Corrected in fee2f3d. All inaccurate OADP and cert-manager citations have been removed or corrected throughout the document:

- This line no longer claims OADP uses bundle-shipped CredentialRequests
- The Summary no longer references "OADP and cert-manager" as CCO-integrated operators
- The "Why the Operator Manages Credentials" section no longer cites "OADP, cert-manager, etc." as using the same pattern
- The Open Design Questions table now accurately attributes the direct Secret creation approach to OADP (`pkg/credentials/stsflow/stsflow.go`) as a distinct third option

PR #41, opened 2026-06-02
PROJQUAY-11215: Adding enhancement for Postgres TLS connections for the operator

bcaton85:
## Summary
  
  - Add enhancement proposal for TLS encryption on operator-managed PostgreSQL ([PROJQUAY-11215](https://redhat.atlassian.net/browse/PROJQUAY-11215))
  - Covers both Quay Registry (`postgres`) and Clair (`clairpostgres`) managed database components
  - Enables opt-in TLS via the existing `overrides` mechanism in the `QuayRegistry` CRD 
  
  ## Motivation
  
  Customers in regulated industries (financial services, healthcare, government) require encryption in transit for **all** connections — including those inside the OpenShift SDN. Today, operator-managed PostgreSQL connections are unencrypted,
  forcing these customers to deploy and manage their own external PostgreSQL to meet compliance requirements.
  
  ## What's in this enhancement
  
  - **CRD changes**: New `TLSOverride` struct on the `postgres` and `clairpostgres` component overrides (`overrides.tls.enabled`, `overrides.tls.secretRef`)
  - **Self-signed certificates**: Operator generates ECDSA P-256 CA + server certs (10-year validity) when no `secretRef` is provided
  - **User-provided certificates**: Support for custom certs via a referenced Secret, compatible with cert-manager output format
  - **Certificate validation**: Reconcile-time validation of user-provided certs with specific error reporting via status conditions and Kubernetes Events
  - **Connection string updates**: `DB_URI` (Quay) and `connstring` (Clair) are updated with `sslmode=verify-full` using proper URL parsing
  - **Migration support**: Enable/disable TLS on existing deployments without data loss 
  - **Certificate mismatch recovery**: Coordinated pod restart when managed keys Secret is deleted or modified
  - **Mixed-mode support**: TLS can be independently enabled/disabled per component
  
  ## Non-Goals
  
  - Mutual TLS (mTLS) / client certificate authentication
  - Automatic certificate rotation (users needing this should use cert-manager + `secretRef`)
  - Direct cert-manager CRD integration
  - TLS for Redis or external PostgreSQL
  
  ## Test Plan
  
  - [ ] Unit tests: CRD validation, cert generation, middleware injection, connection string parsing, `secretRef` validation, CA mismatch detection
  - [ ] E2E tests (Chainsaw): full deployment with TLS, `secretRef` path, migration/downgrade, mixed-mode
  - [ ] Manual verification: `SHOW ssl`, `openssl s_client`, connection tests
  
  ## Related
  
  - Jira: [PROJQUAY-11215](https://issues.redhat.com/browse/PROJQUAY-11215)
  - See also: `enhancements/tls-managed-component.md` (external HTTPS — separate concern)

### DIFF
--- a/enhancements/tls-managed-postgres.md
+++ b/enhancements/tls-managed-postgres.md
@@ -1,0 +1,455 @@
+---
+title: TLS for Operator-Managed PostgreSQL
+authors:
+  - "@bcaton"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2026-06-02
+last-updated: 2026-06-02
+status: implementable
+see-also:
+  - "/enhancements/tls-managed-component.md"
+---
+
+# tls-managed-postgres
+
+Enable TLS encryption for operator-managed PostgreSQL deployments used by
+Quay Registry and Clair, satisfying regulatory requirements for encryption
+in transit within the cluster network.
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Open Questions
+
+> 1. Should certificate rotation be automated for self-signed certs, or is
+>    manual rotation (delete managed keys entry, trigger reconcile) acceptable
+>    for the initial implementation?
+> 2. Should the operator emit Kubernetes Events for TLS-related state changes
+>    (cert generated, cert rotation, TLS errors)?
+
+## Summary
+
+The Quay Operator deploys managed PostgreSQL instances for both Quay Registry
+and Clair as a convenience option. Currently, all database connections are
+unencrypted. Customers in regulated industries (financial services, healthcare,
+government) require encryption in transit for **all** connections, including
+those inside the OpenShift SDN. This is a regulatory and audit requirement that
+blocks deployment of Quay to their platforms.
+
+This enhancement adds opt-in TLS support for operator-managed PostgreSQL via
+the existing component override mechanism in the `QuayRegistry` CRD:
+
+- PostgreSQL pods are configured to require TLS connections.
+- Quay and Clair `DB_URI`/connection strings are automatically updated with
+  TLS parameters (`sslmode=verify-ca`).
+- Certificate provisioning supports both operator-generated self-signed
+  certificates and user-provided certificates (including those managed by
+  cert-manager).
+
+This is distinct from the existing `tls` managed component
+(`tls-managed-component.md`), which handles Quay's external HTTPS endpoint.
+This enhancement addresses internal database transport encryption.
+
+## Motivation
+
+Customers in regulated industries such as financial services, healthcare,
+and government require encryption for all traffic, including internal
+communication within the OpenShift SDN. This is a regulatory and audit
+requirement that blocks deployment of workloads to their platforms and
+hinders adoption (RFE-7835, RFE-9068).
+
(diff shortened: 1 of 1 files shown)


enhancements/tls-managed-postgres.md
```diff
@@ -0,0 +1,455 @@
...
+kind: QuayRegistry
+metadata:
+  name: my-registry
+spec:
+  components:
+  - kind: postgres
+    managed: true
+    overrides:
+      tls:
+        enabled: true
```
jbpratt, line 152: @bcaton85 need to think on it a bit but this seems sort of awkward/confusing with having a tls component already. Not sure if there is a better name for this

bcaton85: yeah i could see it either way. IMO this is more clear because it's explicitly the TLS for the postgres components. @tlwu2013 feel free to chime in on which flow users would prefer

tlwu2013: thanks for tagging me (just saw this now...)

yeah, I won't be worried about naming it as `overrides.tls` here since it's under the postgres component vs `kind: tls` as a top-level component.  And that aligns with the K8s convention quite well (e.g., both Ingress and Gateway API uses tls).

I noticed this is also explicitly called out in the Summary:
> This is distinct from the existing tls managed component (tls-managed-component.md), which handles Quay's external HTTPS endpoint. This enhancement addresses internal database transport encryption.

so once that also got carried into user documentation, I think that'll also help.

jbpratt: > And that aligns with the K8s convention quite well (e.g., both Ingress and Gateway API uses tls)

Yeah this makes me lean towards in a v2 CRD I would want to name the `tls` component `externalTLS` or something explicit rather than 2 things named `tls` doing the different things - our existing CRD is a mess anyway - alright sounds good!

enhancements/tls-managed-postgres.md
```diff
@@ -0,0 +1,738 @@
...
+This enhancement addresses internal database transport encryption.
+
+## Motivation
+
+Customers in regulated industries such as financial services, healthcare,
+and government require encryption for all traffic, including internal
+communication within the OpenShift SDN. This is a regulatory and audit
+requirement that blocks deployment of workloads to their platforms and
+hinders adoption (RFE-7835, RFE-9068).
+
```
ibazulic, line 85: Customers in general should not view the Quay operator deployed database to be production ready. It is not HA and there's no guarantee that the db will self-heal if something goes wrong. We should change our paradigm and should tell people explicitly to deploy their own database solutions rather than providing a system that is inadequate. Or, provide a system that is adequate OOTB.

enhancements/tls-managed-postgres.md
```diff
@@ -0,0 +1,738 @@
...
+Quay already supports TLS connections to **external** PostgreSQL via `DB_URI`
+configuration (e.g., `?sslmode=verify-full&sslrootcert=/path/to/ca.crt`).
+However, when using the operator-managed PostgreSQL convenience deployment,
+there is no way to enable TLS. Customers who require encryption in transit are
+forced to deploy and manage their own external PostgreSQL, significantly
+increasing operational burden.
```
ibazulic, line 91: Which is exactly what they should do, given that this is not a HA solution.

enhancements/tls-managed-postgres.md
ibazulic: Just don't do it. And instead say that default operator deployments for PostgreSQL do not satisfy certain requirements because they are little more than POCs. And what we recommend is a *fully capable HA database solution* provided by an external contractor.

PR #42, opened 2026-06-12
QUAYIO-1637: docs: add spam detection enhancement

HammerMeetNail:
## Summary

Adds an enhancement proposal for [QUAYIO-1637](https://redhat.atlassian.net/browse/QUAYIO-1637): detecting and remediating spam repositories in Quay without adding new in-tree Quay UI.

The proposal defines a service-tool-centered design:
- `quay-service-tool` owns classifier configuration, action policy, training examples, scan execution, scan history, match records, quarantine state, triage workflows, audit history, and redaction history
- detection uses Bayesian filtering rather than keyword-list rules
- service-tool preview and automated scans read repository descriptions and supporting metadata from a read-only Quay database replica
- service-tool scan matches and quarantine-review eligibility require an empty repository with no pushed image content, even when the classifier score exceeds the scan threshold
- service-tool preview and scan results include hard-filter results so operators can tune objective false-positive controls such as visibility, namespace exemptions, account/repo age, URL count, push/build history, and collaboration signals
- service-tool dry-run scans persist run/match history in service-tool state without changing Quay repository content
- quarantine state is stored in the service tool, while approved quarantine, restore, and redaction actions make narrow, explicit writes to Quay repository content
- approved quarantine replaces the repository description with a standard notice that includes restore contact details, expected owner remediation, and review timelines
- Quay in-tree work is limited to repository-description ingress evaluation/enforcement and the configuration needed to load the baked classifier artifact

It also defines the required `quay-service-tool` work as the operator interface:
- classifier and policy management
- training example import and curation
- preview of saved classifiers or unsaved policy drafts
- historical run and match triage
- triage queue actions for quarantine, restore, dismiss, and redaction
- concrete Flask-RESTful backend resources, React/PatternFly route structure, role plumbing, and DB access constraints based on the existing service-tool architecture

Key design decisions called out:
- spam detection state lives in the service tool, not new Quay application tables
- preview is ad hoc and non-persistent; dry-run scans persist service-tool history but do not mutate Quay content
- empty-repository status is a mandatory service-tool scan eligibility gate, not merely a weighted classifier feature
- hard identifiers are objective repository or namespace predicates evaluated outside the Bayesian score and stored in policy snapshots
- OpenShift CronJob invoking the service-tool scan entry point is preferred for automated scans
- multiple Quay pods do not duplicate scans because Quay no longer owns the scanner
- all Quay pods must use the same classifier/policy version for ingress enforcement
- ingress blocking uses `FEATURE_SPAM_DETECTION` and `SPAM_DETECTION_DRY_RUN`
- Quay ingress loads a versioned JSON classifier artifact baked into the image at `/conf/spam-detection/classifier.json`
- classifier updates require exporting a new JSON artifact, rebuilding the Quay image, and rolling all Quay pods to the same artifact version
- initial Quay ingress classifier support uses a fixed tokenizer policy; arbitrary custom regex support is deferred until there is a dedicated safety design
- Quay production rollback is forward-only: disable the feature and preserve service-tool state for forward recovery


HammerMeetNail:
Proposed refinements to the enhancement:

## Detection Approach

- Move away from keyword-based detection in favor of **Bayesian filtering**. This should be applied both **at ingress** (to catch spam at creation time) and **on scan** (to detect existing spam content). A statistical approach avoids the cat-and-mouse game of maintaining keyword lists.

## Service Tool as the Central Orchestrator

- All database-related changes for spam detection should be handled by the **service tool**.
- The service tool should be **augmented to handle rules** for spam classification and action.
- The service tool's only interaction with the Quay DB should be:
  1. **Reading** repository descriptions (for analysis/classification)
  2. **Placing** flagged repositories into quarantine

## Quarantine vs. Redaction

- **Quarantine state** will be maintained in the service tool (not in the Quay DB).
- **Redaction** (permanent removal/cleanup) will happen directly in the Quay DB.

This separation keeps the spam detection logic and quarantine management isolated in the service tool, while only the final redaction step touches the Quay database directly.

HammerMeetNail:
Related implementation PRs:
- Quay ingress classifier support: https://github.com/quay/quay/pull/6154
- Service-tool training, scanning, quarantine, and remediation workflows: https://github.com/quay/quay-service-tool/pull/216

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+
+* that the repository description was removed because automated spam detection
+  flagged it for review;
+* the support contact or restoration request URL;
+* the remediation expected from the owner, such as removing promotional,
+  deceptive, or unrelated link content before requesting restore;
+* the expected review timeline after a restore request is filed; and
+* a stable reference to the namespace/repository and quarantine record where
+  that can be included without exposing sensitive internal data.
+
```
deshpandevlab, line 444: The quarantine notice mentions including "a stable reference to the quarantine record." The implementation should explicitly prevent exposing the internal quarantine record UUID in the public-facing repository description — it could be used to enumerate quarantine activity.

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+| `created_at`, `updated_at` | Timestamps |
+| `created_by`, `updated_by` | Operator identity where available |
+
+Training examples should also be service-tool owned. At minimum, each example
+needs bounded repository text, label (`spam` or `ham`), source
+(`manual_review`, `quarantine`, `restore`, `dismiss`, or import), operator
+identity where available, and timestamps. The service tool should support model
+retraining or model refresh from those examples without requiring a Quay
+database migration.
+
```
deshpandevlab, line 174: Should the implementation enforce a minimum corpus size before allowing a model to be trained or exported? A model trained on too few or imbalanced examples could silently degrade. Consider requiring minimum example counts and surfacing basic quality metrics (precision/recall) at export time.

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+per-repository database queries. Repositories that are not empty at scan time
+must be excluded before writing match history, opening quarantine-review records,
+or applying quarantine actions, even when the classifier score exceeds the scan
+threshold.
+
+The read path should use a Quay database account or replica that is read-only at
+the database permission layer. The service-tool scanner and preview helpers
+should also enable a read-only session where the selected database supports it,
+so accidental writes through the scan connection fail early.
+
```
deshpandevlab, line 480: Potential race condition: The empty-repository gate uses the read-only replica. If a spammer pushes a tag within the replica lag window, the scanner may see the repo as non-empty and skip it. Is the accepted stance that ingress is the primary defense and the scanner is best-effort for this window, or should the emptiness check hit the primary for candidates that pass the classifier threshold?

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+| `SPAM_DETECTION_CLASSIFIER_SHA256` | unset | Quay | Optional SHA-256 checksum for the local classifier artifact |
+| `SPAM_DETECTION_FAIL_OPEN` | `true` | Quay | Allows repository updates if the ingress classifier is unavailable |
+| `SPAM_DETECTION_READONLY_DB_URI` | unset | service tool | Read-only Quay replica used for preview and scans |
+| `SPAM_DETECTION_WRITE_DB_URI` | unset | service tool | Write-capable Quay DB path for approved quarantine, restore, and redaction |
+| `SPAM_DETECTION_QUARANTINE_DESCRIPTION` | deployment-provided notice | service tool | Standard quarantine notice written by approved quarantine actions |
+| `SPAM_DETECTION_BATCH_SIZE` | `200` | service tool | Repositories scanned per batch |
+| `SPAM_DETECTION_SLEEP_BETWEEN_BATCHES` | `0.5` | service tool | Delay between scan batches |
+| `SPAM_DETECTION_SCAN_DRY_RUN` | `true` | service tool | Report matches without opening quarantine records or changing Quay content |
+| `SPAM_DETECTION_MAX_REPOS` | `0` | service tool | Max repositories per scan, where `0` means unlimited |
+
```
deshpandevlab, line 577: Should this enhancement also document the expected JSON schema of classifier.json (required fields, types), or is that intentionally deferred to the implementation PRs? Stating this explicitly would set expectations for reviewers.

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+artifact download, shared mutable volume, or service-tool call from Quay pods.
+
+Ingress checks should evaluate the proposed repository description and other
+request-local fields available on the create/update path. When enforcement is
+enabled, rejections should return a clear validation error. Quay should avoid
+persisting spam-specific state for ingress decisions in new Quay tables.
+Operational visibility for classifier behavior should come from service-tool
+preview, scan history, and standard Quay logs or metrics.
+
+### Audit Logging and Notifications
```
deshpandevlab, line 632: Major: Can we add specific metrics for ingress spam evaluation? Something like:

Evaluation counts by result (pass/block/dry_run/error)
Fail-open counter for requests processed without classifier verification
Evaluation latency on the request path
A fail-open counter would be especially nice to have — it would surface a broken or missing artifact between image builds that might otherwise go unnoticed.

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+  descriptions and supporting metadata.
+* Use `quay-service-tool` as the operator interface for classifier training,
+  threshold configuration, action policy, preview, scan execution, run history,
+  quarantine review, and redaction.
+* Keep spam-detection configuration and quarantine state out of the Quay
+  application database schema.
+* Limit service-tool writes to the Quay database to explicit repository
+  quarantine, restore, and redaction actions.
+* Require scan candidates to have no pushed image content before they can be
+  treated as spam matches or opened for quarantine review.
```
lechuk47, line 62: A repository description can be edited after an image push, so will that prevent a repo for being scaned?

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+  repositories would match a classifier and filter before any Quay data is
+  changed.
+* Support dry-run automated scans that persist service-tool run and match
+  history without quarantining or redacting repositories.
+* Store scan/run metadata in the service tool so operators can answer which run
+  matched a repository and how many repositories were flagged.
+* Allow administrators to quarantine, restore, dismiss, or redact suspected
+  spam through the service tool.
+* Make quarantine reversible by preserving original repository content in
+  service-tool-owned quarantine records.
```
lechuk47, line 73: Is it possible to define the responsabilities that all this adds to the Quay.io team?

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+* Apply the Bayesian classifier at repository-description ingress so Quay can
+  reject spammy create/update requests when the feature is enabled for
+  enforcement.
```
lechuk47, line 76: I don't know much about ML, how long will this take to generate a result?

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+* Use a read-only database replica for service-tool preview, scanning, and
+  historical reporting workflows.
```
lechuk47, line 78: So this will use a different schema on the Quay.io database instance?

enhancements/spam-detection-and-quarantine.md
```diff
@@ -0,0 +1,845 @@
...
+| `training_corpus_version` | Version of the spam/ham examples used to train the model |
+| `model_snapshot` | Serialized model parameters or a pointer to a stored model artifact |
+| `feature_config` | JSON options for tokenization and metadata features |
+| `scan_threshold` | Probability or score required to flag a repository during scans |
+| `ingress_threshold` | Default ingress threshold used when the classifier is trained or exported outside the active policy |
+| `created_at`, `updated_at` | Timestamps |
+| `created_by`, `updated_by` | Operator identity where available |
+
+Training examples should also be service-tool owned. At minimum, each example
+needs bounded repository text, label (`spam` or `ham`), source
```
lechuk47, line 169: Where does the models/training data come from?

enhancements/spam-detection-and-quarantine.md
HammerMeetNail: Yes, ingress is primary defense and scanner is future cleanup.

enhancements/spam-detection-and-quarantine.md
HammerMeetNail: Can we earmark this for a follow up PR?

enhancements/spam-detection-and-quarantine.md
HammerMeetNail: Yes, image content is a key indicator of positive use. It might be something that we remove in the future, but seems like a good initial safety check to avoid false positives.

enhancements/spam-detection-and-quarantine.md
HammerMeetNail: Less than a minute

enhancements/spam-detection-and-quarantine.md
HammerMeetNail: Agreed. I updated the enhancement to make the classifier artifact contract explicit in the ingress-blocking section. I kept this as the required JSON object contract rather than requiring a separate formal JSON Schema for the initial implementation, and the implementation PRs now treat that field list as the contract: service-tool emits it and Quay validates required fields/types before loading the artifact.

enhancements/spam-detection-and-quarantine.md
HammerMeetNail: Agreed. I updated the proposal and service-tool PR so export is gated by configurable minimum spam and ham example counts. Artifacts now include training_metrics with corpus size, spam/ham counts, class balance, minimum thresholds, and validation status. Precision/recall are represented but marked not_available until we add a validation split or labeled holdout; the enhancement now calls out surfacing those metrics when available rather than implying quality was measured when it was not.

enhancements/spam-detection-and-quarantine.md
HammerMeetNail: Agreed. I updated the restoration notice section to make that explicit: the public repository description must not expose the internal service-tool quarantine record UUID or any enumerable internal quarantine identifier. The public notice can include namespace/repository and an external support case or restoration request reference; the internal UUID stays in service-tool state, operator UI, and audit history only.

deshpandevlab (approved):

PR #45, opened 2026-08-11
Add workload identity authentication enhancement proposal

Marcusk19:
## Summary

- Add an enhancement proposal for workload identity authentication for the Quay Management API.
- Document the Kubernetes-to-Quay trust boundary, token validation, token issuance, compatibility, rollout, and security considerations.

## References

- [PROJQUAY-12483](https://redhat.atlassian.net/browse/PROJQUAY-12483): https://redhat.atlassian.net/browse/PROJQUAY-12483
- [PROJQUAY-11090](https://redhat.atlassian.net/browse/PROJQUAY-11090): https://redhat.atlassian.net/browse/PROJQUAY-11090

### DIFF
--- a/enhancements/workload-identity-authentication-for-quay-management-api.md
+++ b/enhancements/workload-identity-authentication-for-quay-management-api.md
@@ -1,0 +1,141 @@
+---
+title: workload-identity-authentication-for-quay-management-api
+authors:
+  - "@Marcusk19"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2026-08-04
+last-updated: 2026-08-04
+status: provisional
+see-also:
+  - "https://redhat.atlassian.net/browse/PROJQUAY-12483"
+  - "https://redhat.atlassian.net/browse/PROJQUAY-11090"
+---
+
+# Workload Identity Authentication for Quay Management API
+
+
+## Status
+
+*First draft for architecture review. This document captures the proposed boundaries and unresolved decisions; it is not an implementation plan.*
+
+## 1. Purpose
+
+Enable Kubernetes workloads to obtain scoped Quay Management API OAuth tokens using their ServiceAccount identity, without storing a shared Quay administrative credential.
+
+## 2. Goals and non-goals
+
+#### Goals
+
+- Authenticate eligible Kubernetes ServiceAccounts to Quay.
+- Issue standard Quay OAuth tokens with enforced scopes and normal expiration/revocation behavior.
+- Preserve existing human, non-Kubernetes, and programmatic-bootstrap flows.
+- Provide auditable success and failure outcomes.
+
+#### Non-goals
+
+- Replacing human authentication.
+- Removing programmatic bootstrap in this feature.
+- Defining class-level implementation details.
+
+## 3. End-to-end architecture
+
+Workload → Quay Management API: presents a ServiceAccount bearer token and requested access.
+
+Quay → Kubernetes API: validates the token through TokenReview (provisional recommendation).
+
+Quay: resolves the validated identity, applies the selected Quay authorization mapping, and reuses the existing organization token endpoint/lifecycle.
+
+Quay → Workload: returns a scoped OAuth bearer token, or rejects the request without issuing one.
+
+*Trust boundary: a valid Kubernetes identity is not, by itself, Quay authorization. Quay must apply an explicit authorization decision before minting a token.*
+
+## 4. Validation approach
+
+*Working recommendation:* Kubernetes API validation using TokenReview, followed by Quay-side authorization. This keeps Kubernetes as the authority for ServiceAccount token validity and rotation.
+
+*Alternative for review:* locally validate the JWT using issuer, signature, audience, and claims. This may reduce API dependency but makes issuer/key-discovery, revocation semantics, and trust configuration Quay concerns.
+
+## 5. Authorization mapping — open decision
+
+How should a validated identity map to a Quay subject and allowed Management API scopes?
+
+- Explicit configured mapping: cluster/namespace/ServiceAccount → Quay identity + scopes. Strong auditability; more configuration.
+- Namespace-based defaults: convenient conventions; risk of unintended access.
+- Kubernetes RBAC-driven mapping: aligns with Kubernetes permissions; scope translation is complex and may couple systems.
+- Quay identity per ServiceAccount: clear attribution; introduces identity lifecycle and cleanup questions.
+
+### 5.1 Authorization decision: SAR or explicit mapping
+
+The design must choose how a validated Kubernetes ServiceAccount becomes authorized for Quay access.
+
+#### Option A — Kubernetes SubjectAccessReview (SAR)
+
+Quay validates the ServiceAccount with TokenReview, then asks Kubernetes SAR whether the identity may perform a defined action. This makes Kubernetes RBAC the authorization authority.
+
+**Advantages:** reuses Kubernetes RBAC and avoids introducing a second mapping model.
+
+**Concerns:** Kubernetes verbs/resources do not map naturally to Quay repository, organization, and **Management API** scopes; it adds another Kubernetes API dependency and call; and it makes Quay permissions indirect and harder to audit. **SAR** is appropriate only if the product deliberately defines Kubernetes RBAC as the source of truth for Quay access.
+
+#### Option B — Explicit administrator-controlled mapping (preferred working direction)
+
+Quay validates the ServiceAccount with TokenReview, then looks up an administrator-created mapping from cluster identity, namespace, and ServiceAccount to a Quay subject and an allow-list of Quay scopes.
+
+**The mapping must explicitly define:**
+
+- Which cluster or Kubernetes issuer is trusted.
+- Namespace and ServiceAccount identity (including audience/issuer constraints where required).
+- The Quay subject to use, such as a robot account or service identity.
+- The permitted Quay organizations, repositories, operations, or Management API scopes.
+- Lifecycle, ownership, audit, and revocation behavior.
+
+#### What is the mapping object?
+
+Preferred for Operator-managed deployments: a dedicated Kubernetes CRD, for example WorkloadIdentityMapping, managed declaratively by the Quay Operator. The Operator validates the object and delivers a normalized mapping to Quay; Quay remains responsible for token issuance and Quay-scope enforcement.
+
+**Other representations to evaluate:**
+
+- Quay-native configuration values: suitable for standalone or non-Kubernetes Quay, but less GitOps- and namespace-native.
+- A ConfigMap: easy to deploy, but weakly typed and not ideal for validation or lifecycle status.
+- A Secret: not appropriate for the mapping itself; it may hold credentials but should not be the policy interface.
+- An administrative Quay API/UI: useful for standalone administration, but requires a separate Kubernetes integration path.
+
+#### Decision guidance
+
+Use SAR only if Kubernetes RBAC is intentionally the policy authority. Otherwise, use an **explicit mapping**, preferably a Quay Operator-managed CRD for Kubernetes deployments plus an equivalent Quay-native configuration mechanism for standalone deployments.
+
+## 6. Token issuance
+
+*Working assumption:* reuse the existing organization token CRUD/issuance endpoint and OAuth lifecycle. Workload identity changes how the caller is authenticated; it does not create a second token format or lifecycle.
+
+## 7. Compatibility and rollout
+
+- Additive feature, gated by FEATURE_KUBERNETES_SA_BOOTSTRAP and off by default.
+- Existing authentication and programmatic bootstrap remain functional when workload identity is disabled.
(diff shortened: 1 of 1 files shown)


enhancements/workload-identity-authentication-for-quay-management-api.md
```diff
@@ -0,0 +1,173 @@
...
+
+## 6. Token issuance
+
+*Working assumption:* reuse the existing organization token CRUD/issuance endpoint and OAuth
+lifecycle. Workload identity changes how the caller is authenticated; it does not create a second
+token format or lifecycle.
+
+## 7. Compatibility and rollout
+
+- Additive feature, gated by FEATURE_KUBERNETES_SA_BOOTSTRAP and off by default.
```
bcaton85, line 180: The JIRA (PROJQUAY-11090) requires the output to be "a standard OAuth Bearer token" with enforced scopes. This section says "reuse existing organization token CRUD" but doesn't specify the actual mechanism:

1. Does the SA present its JWT to the existing org token endpoint and receive a scoped OAuth token back? If so, what request parameters does it pass (org, scope)?
2. Or does SA authentication itself produce an identity (e.g., a robot account) that then uses the existing token CRUD APIs separately?

The initial implementation chose option 2 — SA JWT maps to a robot account directly, with no scoped OAuth token issuance. This means the SA gets full robot permissions rather than a scoped token, which doesn't match the JIRA's acceptance criteria or the spec's own §1 ("obtain scoped Quay Management API OAuth tokens").

**Recommendation:** specify the request/response contract. At minimum: what endpoint does the workload call, what does it send (JWT + requested scope), and what does it receive (scoped OAuth token with expiry)?

enhancements/workload-identity-authentication-for-quay-management-api.md
```diff
@@ -0,0 +1,173 @@
...
+- Preserve token expiration, revocation, and audit behavior.
+- Record successful issuance and rejected attempts with enough identity and reason context for
+  investigation, without logging bearer tokens.
+- Define replay and audience requirements for presented ServiceAccount tokens.
+
+## 9. Open questions
+
+- Which authorization mapping model should be adopted?
+- Is SubjectAccessReview required in addition to TokenReview, and what exact decision does it provide?
+- Where should trust and mapping configuration live, and how is it managed by the Quay Operator?
```
bcaton85, line 224: This bullet identifies audience requirements as needed but doesn't define them. The initial implementation sets `verify_aud: False`, disabling audience validation entirely. This means any valid SA JWT from the cluster authenticates to Quay regardless of its intended audience — a token minted for Vault, ArgoCD, or any other service would also work against Quay.

**Recommendation:** define a required audience value (e.g., `SERVER_HOSTNAME` or a dedicated audience like `quay-management-api`). The workload would request a projected token with that audience:
```yaml
volumes:
  - name: quay-token
    projected:
      sources:
        - serviceAccountToken:
            audience: quay.example.com
```
Quay rejects tokens without the expected `aud` claim. This is how IRSA, GCP Workload Identity, and Vault K8s auth all work — audience binding is the primary defense against token reuse across services.

enhancements/workload-identity-authentication-for-quay-management-api.md
```diff
@@ -0,0 +1,173 @@
...
+ServiceAccount identity, without storing a shared Quay administrative credential.
+
+## 2. Goals and non-goals
+
+#### Goals
+
+- Authenticate eligible Kubernetes ServiceAccounts to Quay.
+- Issue standard Quay OAuth tokens with enforced scopes and normal expiration/revocation behavior.
+- Preserve existing human, non-Kubernetes, and programmatic-bootstrap flows.
+- Provide auditable success and failure outcomes.
```
bcaton85, line 37: What's the test plan going to be? how are we going to verify this feature is ready for production?

enhancements/workload-identity-authentication-for-quay-management-api.md
```diff
@@ -0,0 +1,173 @@
...
+from cluster identity, namespace, and ServiceAccount to a Quay subject and an allow-list of Quay
+scopes.
+
+**The mapping must explicitly define:**
+
+- Which cluster or Kubernetes issuer is trusted.
+- Namespace and ServiceAccount identity (including audience/issuer constraints where required).
+- The Quay subject to use, such as a robot account or service identity.
+- The permitted Quay organizations, repositories, operations, or Management API scopes.
+- Lifecycle, ownership, audit, and revocation behavior.
```
bcaton85, line 128: So to be clear, a Quay admin will define the mapping between a k8s SA to an existing robot account?

enhancements/workload-identity-authentication-for-quay-management-api.md
Marcusk19: Good question, I hadn't fully fleshed out that yet but I've added a section on testing so we have an idea of what the e2e behavior should be

enhancements/workload-identity-authentication-for-quay-management-api.md
Marcusk19: seems to be the way it's heading, at least in my head right now, the responsibility for the mapping should fall on the quay admin and we just provide the tools for it to happen

enhancements/workload-identity-authentication-for-quay-management-api.md
Marcusk19: Agreed. The intended flow is ServiceAccount JWT + target organization + requested scopes into the existing organization token issuance path, returning a standard scoped Quay OAuth bearer token with expiration. The direct robot-identity approach was rejected because it gives the workload the robot’s standing permissions rather than a separately scoped token.

enhancements/workload-identity-authentication-for-quay-management-api.md
Marcusk19: Agreed. TokenReview is the primary validation path, with Quay requesting validation for its expected audience; missing or mismatched audiences are rejected. Audience validation is part of establishing a Quay-intended identity, separate from the still-open authorization choice between an administrator-provided mapping and SAR.

enhancements/workload-identity-authentication-for-quay-management-api.md
```diff
@@ -0,0 +1,234 @@
...
+
+Quay validates the ServiceAccount with TokenReview, then asks Kubernetes SAR whether the identity
+may perform a defined action. This makes Kubernetes RBAC the authorization authority.
+
+**Advantages:** reuses Kubernetes RBAC and avoids introducing a second mapping model.
+
+**Concerns:** Kubernetes verbs/resources do not map naturally to Quay repository, organization, and
+**Management API** scopes; it adds another Kubernetes API dependency and call; and it makes Quay
+permissions indirect and harder to audit. **SAR** is appropriate only if the product deliberately
+defines Kubernetes RBAC as the source of truth for Quay access.
```
bcaton85, line 113: The permissions models not lining up properly is a fairly large issue. It's going to be challenging to match RBAC permissions to our existing auth model. It's going to be confusing as the user as well.  It also adds yet another source of truth for authentication model instead of reusing what we have now.

enhancements/workload-identity-authentication-for-quay-management-api.md
```diff
@@ -0,0 +1,312 @@
...
+creation-date: 2026-08-04
+last-updated: 2026-08-19
+status: provisional
+see-also:
+  - "https://redhat.atlassian.net/browse/PROJQUAY-12483"
+  - "https://redhat.atlassian.net/browse/PROJQUAY-11090"
+---
+
+# Workload Identity Authentication for Quay Management API
+
```
tlwu2013: ## 0. Background

Federated robot accounts brought keyless authentication to registry push/pull, but the Management API still requires automation to "masquerade" as a human user. If that human leaves or is deactivated in LDAP/AD, the automation breaks. The common workaround, a "dummy" human user in the identity provider, consumes a seat license, requires bypassing MFA, and violates non-human identity management policies. Robot accounts cannot help here because they are restricted to registry operations and cannot manage the platform itself. This proposal extends the keyless workload identity pattern to the Management API, enabling Kubernetes workloads to authenticate via ServiceAccount identity without stored credentials.

dronenb:
Hey @Marcusk19 — after going through the doc and #6985 in detail, I think this needs to be split into two clearly separate features, not one:

**A. Instance-wide WIF (for Quay-admin-level bootstrapping)**
**B. Org-scoped WIF (for tenant M2M)**

Right now the design conflates these into one thing, and inherits the existing bootstrap feature's ownership model for both. That's the core issue I'd like to see addressed.

**The principle I think is being violated: M2M auth shouldn't need to impersonate a human, even indirectly.** Per section 7 and the schema, every K8s SA exchange — regardless of workload, namespace, or org — mints its token under `BOOTSTRAP_TOKEN_OWNER`, a single shared super-user. Worth being precise about what that account actually is: `quay-admin` (or whatever the configured owner is) is a completely normal, login-capable Quay `User` — same account type as any person, same password-based login, same UI access. It's not a distinct machine-identity type at all, it's just an account that happens to get used this way. That means every workload's auth is only as isolated as that one human-loginable account, and anyone with its credentials (via the UI or otherwise) has the same standing as every M2M caller combined. There's no real separation between "this is a machine" and "this is a person who can type a password into a login form."

The original RFE was asking to move away from faking human identity for automation. This design validates a real, distinct workload identity via the SA JWT, but then the resulting token is stamped with that same loginable admin account rather than the workload itself — so every workload's actions end up audit-logged as "quay-admin" instead of as themselves, and that account remains a standing human-usable credential the whole time. I think this applies at both the instance and org level: neither should resolve to a human-loginable account.

**A. Instance-wide WIF** — this would be the Quay-admin-configured layer: which issuers the deployment trusts, plus the platform-level validation machinery (JWKS caching, signature verification, etc.). This is a reasonable "bootstrapping" concern since it's one-time platform setup done by a Quay admin. Even here though, I'd want the resulting identity to be a dedicated non-human, non-loginable service identity, not a super-user account that can also be used to log into the UI.

**B. Org-scoped WIF** — this is the tenant-facing layer, and I think it's closer to what the original RFE was asking for. Each org should be able to configure its own trust (its own issuers, or a subset of the instance-trusted ones, plus its own claims-to-scope mapping) on its own `OAuthApplication` — similar to how org admins already mint scoped tokens today via `POST /v1/organization/<orgname>/applications/<client_id>/tokens`. The resulting token should resolve to an identity that belongs to that org, not a shared instance-wide account. Orgs should be able to do what human users already can: authenticate as themselves, scoped to themselves.

For both tiers, I'd avoid tying anything back to `BOOTSTRAP_TOKEN_OWNER` or any account capable of interactive login. This needs a real non-human identity concept end to end — robot accounts might be a good starting point, since they're already non-loginable, org-scoped `User` rows — where the identity is traceable back to the actual caller (the org, the workload) and structurally can't be used to log into the UI as a person.

**Separately: I don't think this should be Kubernetes-specific**, in either tier. The RFE named SPIFFE/Spire and K8s ServiceAccounts as two examples of the same underlying ask. As currently designed: `SUBJECT` is hard-regexed to `system:serviceaccount:<ns>:<name>` — no other issuer's token shape could match. JWKS discovery relies on Quay's own mounted K8s credential, which shouldn't be necessary since JWKS is a public endpoint in standard OIDC. `FEATURE_KUBERNETES_SA_BOOTSTRAP` bakes the issuer into the feature name itself.

I'd suggest keeping `ISSUERS` generic and replacing `SUBJECT` with a claims-matcher instead:

```yaml
AUTHORIZED_SUBJECTS:
  - ISSUER: "https://kubernetes.default.svc"
    CLAIMS: {sub: "system:serviceaccount:ci:tekton-sa"}
    SCOPES: "repo:read repo:write"
  - ISSUER: "https://token.actions.githubusercontent.com"
    CLAIMS: {sub: "repo:my-org/my-repo:ref:refs/heads/main"}
    SCOPES: "repo:read"
```

That way Kubernetes ServiceAccounts are the first supported issuer rather than the only one the schema can express.

Marcusk19:
thanks for the feedback @dronenb - re: 'Instance-wide WIF (for Quay-admin-level bootstrapping)' how much of a need do you have for this? what I'm understanding is Org-scoped WIF is what your team is more interested in, so if I'm misunderstanding your use-case by hamfisting in the exiting oauth bootstrap feature (ie. superuser minting tokens) then I'll only focus on the org-scoped level and abandon the idea of re-using the programmatic bootstrap tokens

Marcusk19:
@dronenb I've updated the enhancement doc PTAL if/when you have the chance. tldr; I'm dropping the "instance-wide" WIF from this feature as it sounds like the original RFE is asking more for this "org-wide" WIF.

To accomplish this, I have expanded the already existing implementation for [robot federation](https://redhat.atlassian.net/browse/PROJQUAY-7803) and allowing that token returned in the exchange to be used for quay api operations as well. This should address most of your concerns, namely it's tenant-facing, does not assume a human identity, and is non-kubernetes specific

dronenb:
Hey @Marcusk19, this looks good to me. I think having the instance wide WIF would be useful as a later enhancement, but this looks like it matches with what I'd envision for the more immediate need that we have.

